PR #36, opened 2017-04-05
Pulled out immunity to bulltrue and pre-emptive strike.

Michael-Hedges:
This should fix Coverslide and Combat Roll making them balanced.


trowl223:
We've been discussing this. Go ahead.

PR #37, opened 2017-04-05
Update weapon skill tree, 03_Feats.txt

1sourcecontrol:
Added a 'rough draft' of the proposed changes to the weapon feat tree, removed the old weapon feat tree.

I need more ideas for the "special feat" or special bonus added at the master level.. and what to put for two of the 'skill steps' on most of the weapon classes. I've left them blank. I can potentially just remove those two skill steps.

Also, I've left out bows because idek what to do for that.

### DIFF
--- a/play/PlayerRules/CharacterCreation/03_Feats.txt
+++ b/play/PlayerRules/CharacterCreation/03_Feats.txt
@@ -388,11 +388,11 @@ apply skill points to unlock the rest of the bonuses. Max skill is 40)
 (Bonuses ONLY APPLY to weapons in their respective weapon class)					
 Proficient: Pistols					
 	Skill @				
-	10:	Pistol Proficient. +5m to ranges R1 & R2			
+	10:	Pistol Proficient. Range +10m			
 	15:	-10% reload miss chance			
 	20:	+10% stealth when firing pistols			
 	25:	Pistol Expert. -5% miss chance @ all ranges			
-	30:	+10m to range R3			
+	30:	+5% Reflex when room clearing				
 	40:	Pistol Master. +20% damage; Gain feat "Pistol Quickdraw"
 		(which is feat "quickdraw"  that is only usable with pistols)			
 		*If you have feat "Quickdraw" in addition to "Pistol Quickdraw",
@@ -400,7 +400,7 @@ Proficient: Pistols
 		your pistols after drawing, you must use an action to holster.
 Proficient: Submachine Guns					
 	Skill @				
-	10:	SMG Proficient. +5m to all ranges (R1, R2, & R3)			
+	10:	SMG Proficient. Range +5m			
 	15:	-10% reload miss chance			
 	20:	+5% Reflex when room clearing		
 	25:	SMG Expert. -5% miss chance @ all ranges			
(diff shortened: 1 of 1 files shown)


bleehu:
For shotguns, the rush: "gaint a -10% to miss if you have moved at least half your speed this turn"
For SMGs, the split: "Gain a -10% to miss if you hit a target already this turn and are firing at a different one."
For Long rifles, "beware the wrath of a patient man": shots fired as a held reaction have a -10% to miss.
For carbines: "Make it count", When in single fire mode, +2% chance to double-crit.
For heavy weapons: "Don't Touch Sasha": -25% to miss at R1.
For explosives: "Get Down!": Allies in blast radius may make a DC 60 Reflex save to negate 30 points of damage from friendly fire.

1sourcecontrol:
jExcellent. Thanks for the feedback.

Any ideas for additional passives that I can add to the skill steps?
I'm thinking of just removing 3 steps per weapon tree, including the -10% reload improvement. Since there already is a weapon use skill set. If I did that, I'd combine some improvements for pistols and other.

bleehu:
I think in general, the more granularity the better. If we can avoid removing steps, lets do so.
Suggested Passives:
Reduce chance to miss
Increase chance to single crit
Increase chance to double crit
Increase damage
Reduce chance to jam
Increase range
Increase chance to miss armor
Negate half action to aim
Negate half action of reload time
Negate half action to draw
Increase reflex roll when clearing room
Add penalty to dex rolls when trying to parkour while holding a weapon, then negate it in skill tree.
Increase perception rolls to spot enemies at range
Increase perception checks for awareness with close range weapon
Increase charisma rolls for intimidation with terrifying big gun.

1sourcecontrol:
Let me know what you think. I added/revised a lot of the feats you proposed (I don't want to allow more than a -5% constant miss chance (with some exceptions) via passives at this point because I'm wary of the consequences with other parts of the weapon revisions. [If players have a negative miss chance, then fire automatic weapons, they will be stupid powerful at most ranges])
Also since criticals only apply ~20 damage bonus, I gave it a significant boost for the carbine master skill, over what you proposed. I know there are other issues brought up by having a large critical chance, but I think the bonus is reasonable (or even not strong enough) considering carbines are very 'middle ground' weapons.

bleehu (approved):
I really like this. I also strongly agree with preventing a negative to-miss percentage passively. 

I really like the variety here.

bleehu:
Do we need to clarify somewhere what R1, R2, and R3 are before we merge this?

1sourcecontrol:
Up to you guys, it will be universally plain what the range (R1, R2, etc.) brackets are if the player understands the guns stats, when the gun revisions are implemented. I can make a temporary change to convert all these weapon skill trees to the old range system, but since the old weapon skill tree system isn't in need of a hotfix, we could just leave it this way (unimplemented in the old gun system) and merge all the changes at once when we're done with the gun revisions? I hope that made sense lol.

bleehu:
I think for this small merge, let's make it compatible with the old system. 

Let's create a much larger change branch called "new range system" or some such that contains the R1, R2, etc. system and update all weapons with their range brackets. Then we'll merge this small change now, and then merge the large change when it's ready. I'm hoping to then push new range system into play at the same time we push your cover mechanic.

bleehu:
All of which would be pending Taylor's shipit, of course.

Wildlovelies:
only thing I see that might need limiting is the long rifle mastery, with high enough sniping stats, we have given snipers the twitch feat for long rifles, it does not limit # of times per turn , nor does it limit 360 degree targeting...

1sourcecontrol:
@Wildlovelies  Good catch, I modified it.

trowl223:
Great work guys. Change all references to "Double Crit" to "Extra Crit" tho, to reflect the nature of those crits and the verbage in basic rules.

PR #38, opened 2017-04-05
Sleight of hand combat redo alex l

1sourcecontrol:
I highly suggest these 2 changes in order to keep all weapons from becoming, effectively, semi-automatic capable, for reasons of player choice, and weapon variety.

"Proposed balance change to "Sleight of hand" so that weapons without semi-automatic capability don't become identical to the same.
Allows the user to negate cocking of a weapon by performing another action first (aiming, moving, reloading, hacking, healing, etc.)
Alternative idea: Negate 1 bolt cycle per turn"

### DIFF
--- a/play/PlayerRules/CharacterCreation/03_Feats.txt
+++ b/play/PlayerRules/CharacterCreation/03_Feats.txt
@@ -283,7 +283,7 @@ upgraded by making good decisions based on this skill, rather than by
 upgrading it with skill points.
 
 Sleight of hand: 
-Prerequisites:
+Prerequisites: DEX 6+
 Negate the pump of a shotgun, cocking of a revolver, or cycling of a bolt when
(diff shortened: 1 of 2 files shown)


bleehu:
ayup. This was a good fix. @trowl223 , can we get a shipit?

1sourcecontrol:
@bleehu I made requirement changes to sleight of hand and to straight pull bolt, and I nerfed straight-pull bolt a little more. Reasons are commented on the commits
Please review and let me know what you think?

bleehu (approved):
I like these changes. I think in context of your explanation in the commit log they make a lot of sense.

In future, use separate branches for separate changes so that Taylor's final sign-off doesn't block changes that aren't directly related.

bleehu:
@trowl223, can we get a shipit?

PR #40, opened 2017-04-06
Mh redact classes

bleehu:
A few classes that probably shouldn't be in play are of course there. 

What's the problem with Paladin right now anyway? Do we have an issue to track that?


bleehu:
@trowl223, can I get a shipit?

trowl223:
Paladin needs to have the nanite bonus added (back in?) as requested in Issue #57

PR #41, opened 2017-04-06
Update DoT

1sourcecontrol:
Adjusted damages down on some conditions, made several conditions explicitly damage stackable, added plasma armor reduction. Stackable (albeit, reduced) damage gives more lethality to conditions, balancing the fact that guns are more powerful.

If approved, these changes would need to propagate to relevant files where the effects are restated.

### DIFF
--- a/play/PlayerRules/Basic Rules.txt
+++ b/play/PlayerRules/Basic Rules.txt
@@ -344,7 +344,7 @@ Pinned - Target pinned (hunkered, no movement) for 2 full turns or until they
 pass a Will saving throw at DC of 70, whichever is quicker.
 Plasma Burn - 1d10+1 stackable damage per turn, for 2 turns if the target's armor
 is pierced/missed. If the targets armor is hit/pierced, the armor's AP level is 
-reduced by 1, permanently (repairable), not stackable. 
+reduced by 1, permanently (repairable, DC 60% Repair check), not stackable. 
 Poisoned - d10 damage per turn. After 5 turns, some poisons may induce 
 unconsciousness. DC 50, Fortitude. Greater / Lesser posion increases damage 
 (+5/-5) and DC (+25,-10).
--- a/play/PlayerRules/Basic Rules.txt
+++ b/play/PlayerRules/Basic Rules.txt
@@ -321,9 +321,9 @@ players fail, they may roll each turn at a DC of 70 to wake up. They continune
 to roll either for (15 - FORT) turns, or until they pass.
 
 Conditions:
-Bleeding - d6 stackable damage per turn. Broken with Medicine Skill check. DC varries. 
-If you have a bleed condition, seek medical aid immediately, You CAN Bleed Out,
-you have been warned.
+Bleeding - d10/2+1 (drop the remainder) stackable damage per turn. Broken with 
+Medicine Skill check. DC varries. If you have a bleed condition, seek medical aid
+immediately, You CAN Bleed Out, you have been warned.
 Burning - 2d10 damage per turn. Broken with a Maneuver Skill Check. DC 40. 
 Every 2 turns, the degree of the burn intensifies
 Downed - Your health has fallen below 0. Every turn in this condition, you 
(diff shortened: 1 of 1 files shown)


bleehu:
We need to list somewhere what the repair DC is for repairing plasma burnt armor. I think it's a wonderful addition! Just need to dot our i's and cross the t's. 

I went back and forth a couple of times in my head about reducing the damage on DoT and making it stackable. In the end, I like the predictability that reducing the random factor creates. 

The difference in the average between one old burn and one new burn is negligible; 5 - 3 = 2 damage per turn. When our health pools around level 5 are usually between 90 and 150, this is laughable.  And if you roll a 1, your super-expensive plasma gun is not worth the metal it's built with. 

But with stackable damage, we can have plasma SMGs which do a meager 10 or 15 damage per hit, but apply plasma burn or another DoT effect which stacks. This can be a very effective way to deal with enemies with large health pools but little armor, or enemies that heal over time. This change allows us to pose new and interesting tactical puzzles for our players in combat. As such I believe this change improves the game overall and I recommend that once we've decided on armor repair DCs, that we should merge it.

1sourcecontrol:
I put in a DC 60% repair check. That should allow someone with basic (skill = 10) repair a 50-50 chance of doing it during combat scenarios. 
I suggest increasing it if you don't like the DC. Since that increases plasmas power.

trowl223:
I'm not in agreement about using dice other than d10s, originally the system was built to be played with only one type of die so that players didn't have to keep track of all their different dice and when to use them. I also need to take a look at how armor works before I sign off on this, but I think it looks good from what I can surmise

bleehu:
Even if it's not reasonable to assume that we've got the Chessex d4, d6, d8, 2d10, d20 set (which almost all of our players do) then it's at least pretty reasonable to assume they've at least got a 6 sided die somewhere. 

I'd like to see @wildlovelies check this for breakage before we send it to Taylor for final sign off.

1sourcecontrol:
When I get some time, I'll probably commit a d6 change to d10/2+1 (drop the remainder) which is essential a d6 with less variance.

I spoke with Taylor at the last CX mission. Implicitly, he wanted to have a pair of d10s as the only dice. Now explicitly, I'm fine with this idea since we can manipulate numbers

bleehu (approved):
@trowl223, can we get a shipit?

Wildlovelies:
An easy way to prevent super stacking of plasma is limit fire rates, do not base fire rate off a stat, but you have to remember, dual wielding and the possibility of multiple people using plasma are options. easy fix is make them two handed, that will help limit their ability to rapid stack uncontrollably. as far as too many people using them, weapon use exotic with a decent investment will prevent everyone from using them, 10 for pistol, 20 for smg, 30 for assault, etc or how ever you want to balance those numbers. you could also reward more points in exotic by making 2 different stacking types, meaning pistols and smgs stack duration, while assault and heavy stack damage and refresh duration. thus rewarding dedication. also lets end game plasma have better pronunciation and less redundancy also mixing the two  could create awesome effects for boss kills... just an idea

1sourcecontrol:
@Wildlovelies, is super stacking actually a problem, though? If your flesh gets hit by a gun 5 or 8 times through or around armor, you should be dead by my reckoning. 
By another example, let's say you get hit 3 times by a super tiny plasma pistol that does 30 damage + burn. You take 90 damage because you hate wearing armor and have 3d10+3 plasma burn per turn. On average (5.5 average per d10) you will take 19.5 damage per turn for 2 turns. Which gives you the chance to first aid yourself, the average player would possibly survive the first turn without aid. 
I think the DoT is small enough, compared to the weapon damage, that super stacking isn't an issue.

Wildlovelies:
5 or 8 times is low for most of my characters, smg with high dex and spray and pray feats is looking at atleast 24 shots a turn (3 base from 9 dex, +3 from pray and pray, per half action), granted pistols don't do much, I was trying to prevent myself mainly from making a plasma hose. I am compiling a list of build I believe break the game, also, that smg build can be doubled with dual wield, totaling 48+ shots a turn (I say plus because 12 dex is easily manageable with cyborgs bringing the total to 56 shots per turn)

Wildlovelies:
this also doesn't include double tap feat procs.

1sourcecontrol:
I think what you're describing is _not_ a problem with DoT but a problem with other parts of the game. If you hit a player 15 times through or around armor, they should be dead regardless of DoT.

Sidenote: the changes to the weapon skill tree should help with part of that problem.

Issue #35, opened 2017-03-25
External Augmentation, please remove from play until further testing is done


1sourcecontrol:
There are a couple off issues tied together here, (i.e. Wealthy Sponsor's place in the game) but I'm concerned that external augmentation is far too powerful in its current state. In my opinion, external augmentation makes itself the only best option for most playstyles. It also doesn't seem to fit in any specific class style, any class can have it. 
Case study. 
Player build: let's assume the player is a pilot human.
per 10, str   2, fort  4, int  2, dex 5, cha  9, luck 10, totals 42.
@ level 2, the player takes the 2 feats required to get external augmentation (whatever they are? I can't tell for sure from reading the feats list; can't find a description).
@ level 4, the player takes two feats of wealthy sponsor: Gain LUK * Lvl * 750 credits. Which means, 10x4x750 = 30,000 per feat or 60,000 credits(!!!!!!!);

The player buys the following from external augmentation: Heavy Suit: $10190. 2x shoulder mounted Zurlinsk MMG-5 miniguns 2x$2000. Either advanced suit armor or Shield Gen Class C, lets go with the armor for  $700. Advanced Powerplant, increase subsytem count by 5, $1000. Advanced speed leg enhancement +4m/movement : $4000. (I'm going to pretend that the modular additions description _actually_ says that augmentation tuning counts towards your subsystem limit or I could get a ludicrous number of stat increases) augmentation tuning x 4; $4000 x 4 = $16000, +4 to each augmentation stat the suit provides. Enhanced targeting module, $600, ignore miss chance penalty on all installed weapons. 
I think that's about it, lets see what we have here:
Armour: 100%, 45+20= 65 DR
STR +2 +4 = +6 (the player has 8 STR in the suit)
PER +1 +4 = +5 (the player has 15 PER in the suit, which means a 40% PER bonus)
2x  Zurlinsk MMG-5 (Mounted Minigun)
        100m
        50%
        50 rds, 5mm linked.
        80 dam (On hit, two "hits" occur).
        Reload DC 55%, 2 Full Actions.
-2m movement + 4m = +2m bonus (at 5 dex, we move 7m/full action)
We spent 36,490 so we have 23,510 credits left over for two submachine guns for our arm hardpoints, or pistols or a small space ship. Incredible potential damage output, armor, and maneuverability (we didn't touch all of the possibilities, just start with high dex and you can run incredibly fast) which virtually erases the difference between other existing mid-range classes.
-Alex

[...]

PR #45, opened 2017-04-07
Removed External Augmentation

bleehu:
pending fix to Issue #35 reported by both Alex and Stephen.


trowl223:
Bleehu and I discussed this, is fine.

Issue #63, opened 2017-04-13
Double-Tap needs editing


trowl223:
Currently too powerful with the combined crit expansion and damage expansion.
Suggested change: since the feat is literally expending another bullet, I think that the feat should only affect rolls that fall into Extra Crit range (96-100 normally), and they should do double damage. I also think that this feat is linked to a problem with the crit system in general, which may need to be visited here or in another issue thread.

bleehu:
Do you want @1sourcecontrol to write up a proposal for a fix?

Wildlovelies:
Splitting it into two feats could be an easy fix

trowl223:
@1sourcecontrol, @Wildlovelies and I spoke about this in our play session a week ago and yes, we may make double tap similar to what I mentioned, remove the crit range expansion, and add a new feat which enhances the crit range.

1sourcecontrol:
I added two branches + pull requests, please see them for more discussion on the details. Close?

bleehu:
Close it when we merge Pull #74

PR #74, opened 2017-04-13
DoubleTap; Update 03_Feats.txt

1sourcecontrol:
@trowl223 Changed DoubleTap to remove crit range, and I'm a little fuzzy on what the details are for the revision so I just wrote with regards to another free firing action, see changes.

### DIFF
--- a/play/PlayerRules/CharacterCreation/03_Feats.txt
+++ b/play/PlayerRules/CharacterCreation/03_Feats.txt
@@ -113,8 +113,8 @@ Double Tap:
 Prerequisites:
 Whenever your hit falls into extra-crit range, fire another bullet as a free
 action. The extra bullet does not proc any other effects and cannot be used to
-double tap again. Only Useable with Semi-Auto or Full auto weapons. Not applicable
-to SA Revolvers, Bolt Action, Pump-Action, etc.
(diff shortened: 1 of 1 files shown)


bleehu:
Can we add that this ability can't trigger itself recursively? And maybe that this effect can't be used more than once per turn per gun? So as to prevent using machine pistols to empty their mags?

bleehu:
When we merge and delete, close Issue #63

trowl223:
I do think @1sourcecontrol's idea that only semi-auto weapons should be able to crit should be applied to this feat. I'd like to make a motion to change the text so it reads "Only usable with Semi-Auto weapons. Not usable with . . ."

bleehu:
Motions are for democratic organizations. You're director. You don't make motions, you make directives. You can be polite about it, but a committee would take too long to build CX. You can ask for options and opinions, but at the end of the discussion, Picard says, make it so. Not I make a motion ;)

1sourcecontrol:
Alright, I made the text changes. From what I understand, this feat will allow the player to make another weapon roll, correct? Not just a free automatic hit?

1sourcecontrol:
@trowl223 , does this feat mean that the player will get another free weapon attack roll on an Extra-Crit?
Or does the feat mean that the player does double damage when he rolls an Extra-Crit?

Issue #56, opened 2017-04-11
Luck critical modifier too powerful


Wildlovelies:
It has been brought up luck modifies crits to too high a number, removing crits from luck removes the stat mostly from gameplay, so I propose a middle ground.

Let luck modifier be determined by a *2 instead of *4, let it still modify crits, (10% at lvl 10) then crits are still affected negatively as well by dumping the stat,  but I also proposed we let it add to either reflex or initiative. If you build your character unlucky, they go first less in combat, on the contrary, a lucky person might go first more, but again this modifier would be a bonus of 10 at 10 luck, still under the soldier bonus. and could stack with the soldier bonus.

trowl223:
Initiative would make more sense from a mechanics point of view, but not from a logical one. In my opinion this idea should be tested, I believe it would add usefulness to a stat and make an area of the game more dynamic.

bleehu:
The branch was merged before the change could be made. @1sourcecontrol could you finish up this change and we'll merge it again?

1sourcecontrol:
Per our conversation, @bleehu, I propose that we close this issue. I think we should leave crits the way they are, for now. I see a lot less potential for abuse now that nanite-regen is being redone and automatic weapons can't crit.

bleehu:
I think we should make a cap that says you can't increase crit range beyond a roll of 80+ even in the case of temporary luck, etc. I think Wildlovelies could still abuse crits.

1sourcecontrol:
Why don't we have lovelies try?
My thoughts:
There are two ways, that I know of, that could possibly abuse crits.
1. Criticals cause bonus damage.
               Even if a player got 100% extra-crit on every roll, which is +20 damage constantly, he can't use it during automatic fire. It matters for beating armor DR, so they would be better at dealing with high armor targets; however, the virtual double or triple damage automatic fire can get compared to +20-60% damage, that could be attained by single-fire 100%-crit, makes criticals far more insignificant, imo, for higher HP targets or targets with less than 100% armor coverage.
2. Critical's interactions with other abilities and feats.
               Side note, I haven't studied every ability in the game, mostly engineer and medic, so I can't say for sure (plus my class is Human and I make mistakes lol) but here are all the interactions I know of: Double Tap, First Shot Counts, and Nanite Regen. (I left out Never Tell Me the Odds because it just adds to crit chance).

[...]

PR #75, opened 2017-04-13
New Crit Feat; Update 03_Feats.txt

1sourcecontrol:
Added a feat to increase crit chance by 10% and extra crit chance by 5%.

What should the title be? And/or the description?
Here are a few titles (gems) I came up with:
"Lucky Mother#$%^@" - really good (jk LoL)
"Luckiest man in the whole wide world wow so amaze"
"Right Place, Right Time"
"Perfect Aim"
"All I do is win"
"Enhance my Luck, GO!"
"Lucky Strikes"
"Lucky shot"
"I'm so Lucky, lel kidz get0utTheway"
"Aim Small, Miss Small"

### DIFF
--- a/play/PlayerRules/CharacterCreation/03_Feats.txt
+++ b/play/PlayerRules/CharacterCreation/03_Feats.txt
@@ -192,6 +192,11 @@ an additional +20% o unconventional maneuvers in such as well, but you and
 your vehicle take double damage for the next three turns after using these
 bonuses. (any teammates in such a vehicle do not take double damage)
 
+Honestly tho, what is the title?:
+Prerequisites: None
+Somehow you always aim at the right location, at the right time. 
++10% Crit chance, +5% Extra-Crit chance.
+
 Kata: 
 Prerequisites: Level 10 or higher
 Whenever you land a successful melee hit against someone, you may roll for X 


trowl223:
Maybe 5% and 5%? I think on a crit, a 10% improvement is huge.

I'm a fan of "Aim Small, Miss Small". Anyone for "Never tell me the odds", or is that too heavy a reference?

1sourcecontrol:
Couple thoughts:
It is a feat, so +5%/+5% might be reasonable, but it's mostly a 5% chance to see an increase of 10/20 damage which isn't a large difference in dps. So imo, crit matters more for compounding effects, specifically nanite regen/doubletap. Without other factors, we have an average +1.5 damage taking this feat, or ~+6% damage for a 60 damage gun. if I mathed correctly.

If #56 nerfs luck mod that'll make this feat more significant and simultaneously nerf crit builds, which is a good argument for +10%/+5% imo since we won't have to worry as much about people critting on 50% or something ludicrous.

bleehu:
Crits interact with more than just damage; as per nanite regen. I would tend toward making it small and then boosting it if it tests underpowered.

Never Tell Me The Odds is FANTASTIC!!!!

bleehu:
What's our next step with this one?

1sourcecontrol:
I'm waiting for a response on #77. I'll change the text to "never tell me the odds" (even though, "Enhance my Luck, GO!" is far superior).

trowl223:
Updated, proceed.

trowl223:
I am reserving "Aim Small, Miss Small" for another feat then.

1sourcecontrol:
Merged before I could make the update! D:

PR #76, opened 2017-04-13
FixText in WeaponSkillTree update

1sourcecontrol:
Fixed double crit text


bleehu (approved):
I'm Bleehu and I approved this message. @trowl223 can we get a shipit?

Issue #49, opened 2017-04-11
Misc items description needs to be updated


Wildlovelies:
Currently states: 
Examples of Misc. Items include rope (20m), compass and map, matches, 
flashlight, a single magazine for your gun, flares, etc. These are basically 
items that have an actual use of some sort. $100 each unless otherwise 
specified, 5 free at start.

magazines generally are not $100 each, my opinion it needs updating or removed, feel free delete if you disagree

bleehu:
I totally agree. Will write a patch.

trowl223:
I totally agree as well. This was from the pre-money days of compound_x when players used an Equipment-Point inventory management style to "purchase" items.

bleehu:
You said nice start. Is there more misc items that you want?

PR #82, opened 2017-04-18
New Misc Items

bleehu:
changed verbage on misc items and added some suggested misc items with appropriate prices.
In response to Issue #49


trowl223:
Good start! I think i threw some other radio / comms items into the Items proper, we should talk about communications systems maybe in one of our next chats? Nothing too in-depth.

PR #83, opened 2017-04-18
New Crit Feat Update 03_Feats.txt

1sourcecontrol:
Changed feat name to "never tell me the odds" 
I would like to continue the discussion on 10%/5% crit chance.

We nerfed (and buffed) doubletap so that it doesn't increase crit chance. With crits only applying to semi-automatic weapons now, abuse from nanite regen is significantly reduced. In addition, a feat that only grants a 5%/5% increase in crit chance is pretty minor buff. 
That's why my vote is for 10%/5%

But, we can start small.

### DIFF
--- a/play/PlayerRules/CharacterCreation/03_Feats.txt
+++ b/play/PlayerRules/CharacterCreation/03_Feats.txt
@@ -192,11 +192,6 @@ an additional +20% o unconventional maneuvers in such as well, but you and
 your vehicle take double damage for the next three turns after using these
 bonuses. (any teammates in such a vehicle do not take double damage)
 
-Honestly tho, what is the title?:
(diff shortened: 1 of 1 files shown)


bleehu:
*nerdgasm* Please make this a thing!

Does it make you feel weird if I edit your comments?  It feels like a violation of some internet rule - Alex

trowl223:
Please restrict editing of comments to only those made for ensuring correct use of verbiage.
Example: changing references of "Clips" to "Mags" or correcting the name of a referenced feat, skill, ability, etc.

1sourcecontrol:
@trowl223, I thought it was clear that I was just joking around, I'm sorry about that.

Issue #72, opened 2017-04-13
Pistol Attachment limiting not listed


Wildlovelies:
Nit picky, but I cannot find where it says you can only put x attachments on pistol, or the opposite, cannot attach x to pistol

trowl223:
Will address in attachments document or weapons document.
Pistols may make use of the following attachments:
- Sights with the "Pistol" keyword
- Aiming assistances like Lasers and Flashlights
- Barrel Attachments
- Slings with the "Pistol" keyword
- Ammunition
- Total Conversions with the "Pistol" or "Revolver" keyword, as applicable

Any other concerns?

Wildlovelies:
nope, just they weren't listed. remembered this from doctor build

trowl223:
I've added a response on the merge request thread, I'm going to close this issue since a solution is being worked on.

PR #97, opened 2017-05-18
added pistol limitations

bleehu:
As per Issue #72


bleehu:
Can we also add Wildlovelies as a collaborator so I can assign reviews to him?

trowl223:
We might as well just say that all lower rail attachments are not usable on Pistols, that seems to be the trend. The only attachments that can fit and are feasible on a pistol are lasers and lights anyway, or that impractical knife attachment that has been floating around the internet.

bleehu:
Made that change and added small SMGs to the restriction. You're right, that's a much more succinct rule. :)

Issue #57, opened 2017-04-11
Paladin nanite addition


Wildlovelies:
@trowl223  paladins were mentioned in our last play session not having bonus nanites that were intended.

trowl223:
Bonus should be less than a spell caster's bonus, maybe +75 static, or +45 +5/level.

bleehu:
Well don't write it here! Put it in the rules!!! Or if you're not going to make the change yourself, direct either myself, or @1sourcecontrol or even @Wildlovelies  to make the change.

bleehu:
@Wildlovelies do you want to take a shot at making a branch -> making the fix -> making the pull request -> profiting? Or are you having more fun breaking things?

Wildlovelies:
breaking

bleehu:
lol, alright :) @trowl223 mind if I take it?

trowl223:
Not at all.

PR #98, opened 2017-05-18
added nanite bonus to Paladin

bleehu:
To fix Issue #57


trowl223:
Definitely have a few people test this out and see how the nanite management feels. I'm worried about it being a bit light, which means its probably right where we want it.

PR #102, opened 2017-05-18
take a whole bunch of our changes in RC and publish them to Play!

bleehu:
It's almost time to play with our wonderful hard work! 

Before we check this off, I strongly suggest that @Wildlovelies and @trowl223  check the changes with a fine-toothed comb. I really don't want to publish things that aren't ready for play. I highly recommend looking through the "Files changed" tab.


trowl223:
I know we have a play session coming up, I'll get to combing through this tonight.

1sourcecontrol:
I took a quick look at it, the only thing I noticed was that plasma condition in the basic rules has a note about the new armor (which, as we all know, isn't implemented yet). I think we should just leave it, and if any player asks about it, we can just say it doesn't apply right now.

bleehu:
@trowl223 Can we move forward with this?

PR #105, opened 2017-05-22
extra-crit range as F(x) of luck mod

1sourcecontrol:
Per our conversation, @trowl223,  amended crit chance rules to allow extra-crit range to change at half the rate of crit range (as a function of LUCK mod).


bleehu (approved):
I like it! Make it happen!

Issue #67, opened 2017-04-13
OP Build2: Marksman


Wildlovelies:
-Marksman: both builds can be broken, tech rounds with machine gun pistols and nanite regen..
Multi pin at lvl 10 is scary as hell, pick up to 13+ targets, everytime they do anything you get to shoot them

trowl223:
As mentioned in Issue #58, a potential solution is to allow the marksman feats only be usable with long rifles.
For multi-pin we can limit the number of targets to fire at per turn, but maintain that each pinned target /can/ be fired at, it would be user discretion on which targets are a primary threat rather than just firing at everything that twitches.

Wildlovelies:
also we need to limit direction, makes no sense to pin someone and shoot freely at them if they are behind you.

trowl223:
You say "makes no sense". I say its a class ability which enables them to track any targets they have recently hit. Think about the Medic's heartbeat sensor, is that much different? I do think we should limit how far back "recent" is, I'll come up with a solid edit for the feat that addresses the following issues:
- Too many targets are able to be feasably fired at with a bolt-action rifle
- Limit how long a target is considered "watched"

Any other concerns?

Wildlovelies:
No

bleehu:
Heartbeat simply tells you if an enemy is there, it doesn't grant any reactions to their movement. I agree that 'recent' could be more specific than it is, but the way that it's written that would be at DM discretion. And it probably should only apply to long rifles. Seems like an easy fix. @trowl223 do you still want to handle this personally, or do you want to delegate it?

trowl223:
Since @1sourcecontrol is doing the combat update stuff, I'm going to reassign to him. My computer is still buggered and I need to find time to fix the setup so I can resume direct dev work.

bleehu:
fix merged. Closing issue.

PR #107, opened 2017-05-22
nerfing pin/multi-pin

1sourcecontrol:
Per issue #67 Attempt to balance marksman pin/multi-pin abilities.

Nerfed the bonus accuracy. Redesigned other things.
I tried to reword some text to remove vagueness.

### DIFF
--- a/play/PlayerRules/CharacterCreation/02_Classes.txt
+++ b/play/PlayerRules/CharacterCreation/02_Classes.txt
@@ -482,20 +482,21 @@ Starting Wealth: $1700+100*Luck + 2000
 Class Feats:
 ----------------------
 Pin Target
-Select a target. If the target is fired at by you, the target is Pinned. If the 
-target attempts to move from their pinned cover within a 180 degree field of view 
-infront of you, you may roll a DC 60 sniping skill check to 
-hit them. For the remainder of the encounter, you get a +10% to hit the 
-selected target. If the target dies you may select a new target.
-Only usable with long-rifles. The Pin/Multi-Pin ability can be activated by 
-a pinned target up to 3 times per Marksman player's turn cycle.
+If you fire at a target that is in cover, you can choose to "Pin" the target. Only 1 target may be
+Pinned by the marskman at a time. If the target attempts to move from cover within a 180 degree field
+of view infront of you, you may roll a DC 60 sniping skill check to hit them, as a free action. For the
+remainder of the encounter, you have a +10% to hit the pinned target with a firing action. If the target moves, 
+it is still considered pinned, and if the target dies you may pin a new target. Pin can only be activated by
+a pinned target once per Marksman player's turn. Only usable with long-rifles.
 |
 |
 |
 V
 Multi-Pin
-You may select and pin up to PER/2 (round down) targets at a time (you have to fire at a 
-target to pin it).+30% to hit a Pinned target (instead of +10%). Only usable with long-rifles. 
+You can now Pin a maximum of PER/2 (round down) targets at  time (you still have to fire at a 
+target to pin it). +20% to hit a Pinned target (instead of +10%). The Multi-Pin ability can be 
+activated by a pinned target once, per target, per Marksman player's turn. (I.e., 3 Pinned targets can 
+trigger Multi-Pin once each for a maximum total of 3 triggers before a marskman player's next turn).
 ----------------------
 Tech Rounds
 Spend 10 nanites. Choose from the following list of effects:
--- a/play/PlayerRules/CharacterCreation/02_Classes.txt
+++ b/play/PlayerRules/CharacterCreation/02_Classes.txt
@@ -482,18 +482,20 @@ Starting Wealth: $1700+100*Luck + 2000
 Class Feats:
 ----------------------
 Pin Target
-Select a target. +25% to hit. On hit, target is Pinned. If the target attempts 
-to move from their pinned cover, you may roll a DC 60 sniping skill check to 
+Select a target. If the target is fired at by you, the target is Pinned. If the 
+target attempts to move from their pinned cover within a 180 degree field of view 
+infront of you, you may roll a DC 60 sniping skill check to 
 hit them. For the remainder of the encounter, you get a +10% to hit the 
-selected target. If your rifle has Pin listed as an attribute, then any time 
-the selected target is pinned you may attack them reflexively if they move 
-from their pinned cover. If the target dies you may select a new target.
+selected target. If the target dies you may select a new target.
+Only usable with long-rifles. The Pin/Multi-Pin ability can be activated by 
+a pinned target up to 3 times per Marksman player's turn cycle.
 |
 |
 |
 V
 Multi-Pin
-You may pin up to PER targets at a time. +20% to hit a Pinned target.
+You may select and pin up to PER/2 (round down) targets at a time (you have to fire at a 
+target to pin it).+30% to hit a Pinned target (instead of +10%). Only usable with long-rifles. 
 ----------------------
 Tech Rounds
 Spend 10 nanites. Choose from the following list of effects:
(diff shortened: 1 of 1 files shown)


bleehu (approved):
I like making this available to all long rifles and only long rifles. Along with the requirement that you must fire at the target to pin it.

trowl223:
What do you think of the following changes:
"If the target is fired at by you" -> "If you hit the selected target"
The effect should only trigger when a target is hit.

"The Pin/Multi-Pin ability can be activated by a pinned target up to 3 times per Marksman player's turn cycle" -> "This ability allows a marksman to react to a pinned target once per turn."
This version of the feat only allows one target to be pinned, so it doesn't makes sense that the only pinned target would trigger the marksman's reaction shot more than once per turn.

Also, "Only 1 target may be pinned by the marksman at a time." should be added to the first iteration of this feat.

"You may select and pin up to PER/2 (round down) targets at a time (you have to fire at a target to pin it).+30% to hit a Pinned target (instead of +10%). Only usable with long-rifles. " -> "You may monitor and react to up to PER/2 (round down) pinned targets at a time. +20% to hit a pinned target. Only usable with Long Rifles."
Cleared up some phrasing. A target is only pinned when hit, that is defined in the first section of the feat so we don't have to mention that again. The original +20% replaces any other bonus, this is the intended behavior of all class feats. if class X has a level 5 feat that says "+15% to do Y", and the level 10 variant says "+30% to do Y" that means the total bonus to performing Y is 30%. This phrasing is fairly ambiguous and I apologize for that. Maybe we should change all second-tier class feats so that bonuses are written as "The bonus for doing Y increases to +30%". That is fairly explicit.

Are we also adding in that this ability takes actions off of the marksman's turn, or leaving it as-is? I'm fine with either decision, I thought'd I'd bring it up since we'd discussed it in the past.

1sourcecontrol:
>What do you think of the following changes:
>"If the target is fired at by you" -> "If you hit the selected target"
>The effect should only trigger when a target is hit.

1. Why should the effect only trigger when a target is hit?
My reasoning was that I didn't want to straight nerf marksman, so it gives them an advantage this way while other ways get nerfed. Also realism, 'pinning' sounds analogous to  'suppressing' a target, and that is more about psychological warfare than physical damage. Also, it would even make sense if you just aimed at him, for him to be pinned, since him coming out of cover would be stupid. I suppose you may be considering the first shot to be a maiming shot? Also, most long rifles are bolt action, making it much harder to pin targets if it's an on-hit effect (especially in regards to multi-pin).

> "The Pin/Multi-Pin ability can be activated by a pinned target up to 3 times per Marksman player's turn cycle" -> "This ability allows a marksman to react to a pinned target once per turn."
This version of the feat only allows one target to be pinned, so it doesn't makes sense that the only pinned target would trigger the marksman's reaction shot more than once per turn.

2. Why doesn't it make sense? In my mind, the target could be darting from cover to cover on their turn. If I was trying to kill someone, I would be shooting every chance I get. 

>Also, "Only 1 target may be pinned by the marksman at a time." should be added to the first iteration of this feat.

3. "If the target dies you may select a new target." already says that, but we can make it more explicit if you prefer.

>"You may select and pin up to PER/2 (round down) targets at a time (you have to fire at a target to pin it).+30% to hit a Pinned target (instead of +10%). Only usable with long-rifles. " -> "You may monitor and react to up to PER/2 (round down) pinned targets at a time. +20% to hit a pinned target. Only usable with Long Rifles."
Cleared up some phrasing. A target is only pinned when hit, that is defined in the first section of the feat so we don't have to mention that again. The original +20% replaces any other bonus, this is the intended behavior of all class feats. if class X has a level 5 feat that says "+15% to do Y", and the level 10 variant says "+30% to do Y" that means the total bonus to performing Y is 30%. This phrasing is fairly ambiguous and I apologize for that. Maybe we should change all second-tier class feats so that bonuses are written as "The bonus for doing Y increases to +30%". That is fairly explicit.
4.  Ok.

> Are we also adding in that this ability takes actions off of the marksman's turn, or leaving it as-is? I'm fine with either decision, I thought'd I'd bring it up since we'd discussed it in the past.
5. I put the following in, "The Pin/Multi-Pin ability can be activated by a pinned target up to 3 times per Marksman player's turn cycle" for that reason. I figured that it would be ok to grant the marksman the ability to react without extra expense.

bleehu:
I think this is good discussion!
1) I felt pinned by Alex's nerf monstrosity at nerf wars in past even when not hit by his shots. 
2) If the trigger is the target leaving cover, how would it be possible for the marksman's ability to trigger more than once per turn per target? That also doesn't make sense to me. 
3) We should make that much more explicit. Players sometimes create loopholes when they don't quite understand what's written. We should avoid attempting to communicate with nuance. 
4) Whichever way we communicate it, we need to make it very clear.
5) I still don't understand how a target could proc 3 sniper reaction shots in a single turn.

1sourcecontrol:
@bleehu, 
2. If a target re-enters cover and leaves it again. 
Example: I am pinned and hiding behind light cover in a hallway, I run ahead to then next cover in the hallway, proc'ing marksman Pin. Then I leave from that cover, to a third cover (or i'm just standing in the open, w/e). Proc'ing marksman pin a 2nd time.
5. I tried to make it about multi-pin as well as 'single'-pin in one swoop. The way I meant it to be written is that the Marksman player cannot shoot more than 3 reaction shots per his full turn cycle, no matter how many people he is pinning.

bleehu:
That makes a little more sense. =)

1sourcecontrol:
@trowl223, fyi I won't do (don't want to do) any more work on this until you respond to the recent post(s) =)
(I have plenty of other things to do, so take your time)

trowl223:
So sorry! working on this now.
1. Makes sense.
2. I don't think so. The marksman is attuned to only the precise locations of the target he has pinned. if one leaves cover and is not hit by the reaction shot, the marksman has lost track of them temporarily and must readjust. The time frames seem to be in favor of the marksman too much. three shots in the span of their enemy's turn while they've already had a full turn is like granting them extra actions.
The situation will not always be that someone is moving in a straight line down a hallway towards you as in your given example. In that instance, it may be up to the DM to allow that if the Marksman has a semi-auto rifle since they don't have to readjust, but in a different example this may not be the case. What if the Marksman was providing cover to their squad from a bell-tower positioned 200 meters from combat, and they have to watch a 300m spread of land in front of them? A pinned target moving side-to-side form the Marksman's POV would require considerable time for the marksman to readjust if they miss the first shot as the target leaves cover.
3. My thoughts exactly. An ability of this caliber needs to be pretty explicit.
"5". I still don't think that they should receive an extended amount of extra firing time. At the second tier, I think that with multiple targets the Marksman can have multiple reaction shots (as per my original edit) but at the first tier we're talking about a potential Level 5 ability, its too much out of turn action IMO.

1sourcecontrol:
@trowl223, sounds good.

Please review my latest commit

play/PlayerRules/CharacterCreation/02_Classes.txt
```diff
@@ -482,20 +482,21 @@ Starting Wealth: $1700+100*Luck + 2000
...
-infront of you, you may roll a DC 60 sniping skill check to 
-hit them. For the remainder of the encounter, you get a +10% to hit the 
-selected target. If the target dies you may select a new target.
-Only usable with long-rifles. The Pin/Multi-Pin ability can be activated by 
-a pinned target up to 3 times per Marksman player's turn cycle.
+If you fire at a target that is in cover, you can choose to "Pin" the target. Only 1 target may be
+Pinned by the marskman at a time. If the target attempts to move from cover within a 180 degree field
+of view infront of you, you may roll a DC 60 sniping skill check to hit them, as a free action. For the
+remainder of the encounter, you have a +10% to hit the pinned target with a firing action. If the target moves, 
+it is still considered pinned, and if the target dies you may pin a new target. Pin can only be activated by
```
bleehu: ...if the target dies _or escapes from the battlefield_ you may pick a new target

bleehu (changes requested):
I would make one more addendum. After that, shipit!

Issue #34, opened 2017-03-23
Balance Needs More Dakka


bleehu:
Needs more dakka undertuned with new automatic rifles. Need to add a weapon restriction, so cannot be used with autocannon.

Wildlovelies:
could change the auto cannon to just a cannon, which has to be cycled. thus removing it from empty the mag then use similar verbage.

Wildlovelies:
making it like a revolver, but instead of the hammer, you have to manually cycle the drum

trowl223:
Like a shotgun, perhaps?

bleehu:
Autocannon is fine right now, and will be very carefully balanced by the new weapon work that 1Sourcecontrol is doing anyway. The issue is with Needs more Daka:
https://github.com/trowl223/Compound_X/blob/play/play/PlayerRules/CharacterCreation/02_Classes.txt#L59

It's a whole class feat that lets you fire an automatic rifle at a whopping fire rate equal what it normally can. Since most specialist weapons take a high STR to even carry, and now automatic weapons use STR like SMGs use DEX, there's no benifit from this whole class feat. The solution is to buff the class feat, and then prevent that buff from breaking ULAC.

trowl223:
1Sourcecontrol and I are looking into changing the way automatic weapons work in Combat 2.0. Updates on that under his thread.

1sourcecontrol:
During a mission @bleehu had ad-hoc allowed me to use 6 shots instead of 4 for the first feat (which brings it on-par with automatic rifles @ high STR, 3 shots for 1 half action at 9 STR; if I recall correctly).

I think "needs more dakka" could get reworked regardless of the combat 2.0 changes because it's main use seems to be to make any semi-auto weapon fire like a fully-automatic one. Maybe that's ok? But it's basically like using automatic rifles without the better stats of automatic rifles; or if a specialist uses a minigun (like they're designed to do), there's no point in having "needs more dakka."

bleehu:
merged and completed!

PR #109, opened 2017-05-23
Balance Needs More Dakka

bleehu:
Changed condition "Pinned" to "Pinned Down" to avoid confusing it with the sniper ability.

Updated Specialist Class Feat "Needs More Daka" as per Issue #34


bleehu:
Since I raised this issue, and I wrote the fix, I need at least one other person to review my changes and sign off on them before I push so that I'm not being a dictator.

1sourcecontrol (approved):
Looks good to me. I like the fact that it's a modifier to firing actions and not just a 'make everything fully auto' ability.

Btw, we might have to correct a lot of text in other areas that reference the old "Pinned" condition

bleehu:
As best I can tell, that was the only place the pinned condition was used, but I may have missed a feat or an advanced weapon? @trowl223 can you help us parse which condition/effect was which?

Issue #68, opened 2017-04-13
OP Build3: Soldier


Wildlovelies:
-it'll heal soldier plus full armor: up to 50dr 100% coverage, plus only taking half dmg for everything else...

trowl223:
As discussed, soldier can have a third look taken at it. I'm not comfortable with them and they could use something a little different. Yank for now? I feel that the current difference in builds between Specialist and Soldier is pretty small, so we can direct front-line damage dealers to build as a Specialist in the time being.

bleehu:
Sounds like the fix here is to just take a look at the "It'll heal" feat. Soldier and Specialist are already pretty different; take a look at the way Zach's specialist, Alex's specialist and Gnat's specialist play versus how Derek and Kevin's soldiers play. Maybe we can do some minor tweaking to increase the difference between them, but I think yanking them is way overkill.

trowl223:
Alright.
Verdict: Fix It'll Heal. My recommendation is to allow the full amount of "ignored" damage to be taken at the end of the conflict, instead of half or 1/3. This damage does not have DR applied to it, so if a solider lets 4 or 5 shots "heal" during the course of an encounter, they will likely die at the end of the encounter. We can add some restrictions so that these wounds cannot be healed by normal means or require higher DCs to restore the Soldier to full spec.

bleehu:
Sounds good. Who do you want to write it?

1sourcecontrol:
Am I reading this right? So the soldier takes 1/2 damage until the end of conflict, then basically, instantly dies by taking another full set of damage?

bleehu:
@1sourcecontrol That's correct. Maybe we could do better?

trowl223:
Its the whole adrenaline rush thing on a new level. The soldier gets so amped on trying to finish the conflict that they can ignore crazy amounts of pain and damage to accomplish their goal. Once the adrenaline wears off (after encounter is over) they succumb to the blood loss / system failures. If you guys don't like this mechanic we can look at something else or another way of doing it, I'm not super attached to it.

bleehu:
This has been sitting too long, I'm going to assign myself and take care of it.

bleehu:
Taken care of

PR #115, opened 2017-06-22
changed 'It'll Heal' to 'fight feircely first' class feat for soldier

bleehu:
As a potential fix to Issue #68 For your review.

Since this is a minor issue, I'd like to merge this after one approved review as opposed to waiting for unanimous reviews.

### DIFF
--- a/play/PlayerRules/CharacterCreation/02_Classes.txt
+++ b/play/PlayerRules/CharacterCreation/02_Classes.txt
@@ -34,16 +34,18 @@ Allies must be able to hear and understand you to get the bonus.
 V
 Either 75% + (CHA mod) success or 30% added to next allied rolls.
 ----------------------
-It''ll heal [Min. Requirement of 7 fortitude]
-Make a FORT save at 50% to ignore an enemy damage. Remove one action next turn 
-to stare down your enemy, curse at them hitting you, make a badass remark, 
-etc. All damage taken during the encounter stacks, and half is taken at the 
-end of the encounter, regardless of your condition.
+Fight Feircely First
+The Soldier gets an additional +10 to initiative rolls. 
+In addition, the soldier with this class feat gets a +30% to crit against enemies
+who have not taken a turn yet in combat or whom the soldier has beaten in 
+reflex roll when room clearing until that character takes a turn after losing the 
+reflex roll.
 |
 |
 |
 V
-Fort save is at 75%, or you take 1/3 of damages dealt
+The soldier's crit bonus effect works on any enemy with a lower initiative until
+that enemy has finished their 3rd turn. 
 
 "Specialist" - There is always a need for the type of person willing and able 
 to bring the heaviest of fire down upon their enemies. The Specialist will be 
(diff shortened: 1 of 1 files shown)


play/PlayerRules/CharacterCreation/02_Classes.txt
```diff
@@ -34,16 +34,18 @@ Allies must be able to hear and understand you to get the bonus.
 V
 Either 75% + (CHA mod) success or 30% added to next allied rolls.
 ----------------------
-It''ll heal [Min. Requirement of 7 fortitude]
-Make a FORT save at 50% to ignore an enemy damage. Remove one action next turn 
-to stare down your enemy, curse at them hitting you, make a badass remark, 
-etc. All damage taken during the encounter stacks, and half is taken at the 
-end of the encounter, regardless of your condition.
+Fight Feircely First
```
1sourcecontrol: Spelling: Feircely -> Fiercely

play/PlayerRules/CharacterCreation/02_Classes.txt
```diff
@@ -34,16 +34,18 @@ Allies must be able to hear and understand you to get the bonus.
...
+The Soldier gets an additional +10 to initiative rolls. 
+In addition, the soldier with this class feat gets a +30% to crit against enemies
```
1sourcecontrol: I feel that it would be good to be explicit on whether the crit chance bonus operates by normal crit rules (crit doesn't apply to automatic fire weapons, etc.)

1sourcecontrol (changes requested):

bleehu:
@1sourcecontrol's suggested changes have been made. Though I'm a little uncomfortable with how this separates the class feat from the soldier's signature weapon: the assault rifle. Thoughts?

trowl223:
Maybe with an automatic weapon the first shot that hits can have the crit bonus applied? Or a reduced crit bonus for just the first shot?

1sourcecontrol:
@bleehu, I agree, it would be ideal to include assault rifles.

Two additional ideas I have:
1. Allow the ability to crit with all weapons but add something to the effect of, "If an automatic weapon is used (a weapon that fires more than one shot per action) then crits can only trigger bonus damage; crits cannot trigger other feats or abilities"
2. Don't even use "crits" as they are implemented in CX, i.e., uniquely define the bonus: "When rolling accuracy, per shot fired, if the roll is higher than 70%-crit mod, apply +10 damage."

Of the two i listed, I prefer option 1, it is much more interesting for player choice and scales better into the late game.

_note: I'm not worried about crit bonus damage becoming a problem. If you already hit the target with 3-6 rifle bullets, it usually doesn't matter if the bullets crit._

*edited some text for clarity*

1sourcecontrol (approved):

bleehu:
What if we throw the crit bonus out the window and use a different bonus such as accuracy instead? Or perhaps extra movement for the opening tactical positioning?

trowl223:
I'm approving the accuracy change. We'll test the -30, if it's too strong it should be changed. I think 30% is in line with the original strength of class feats though, so it should be pretty balanced with other classes.

bleehu:
And it only lasts for a couple of turns, which I think justifies some strength.

PR #116, opened 2017-07-03
added placeholders for Power Suit rule files.

bleehu:
I have no suggestion that these chassis, suits or even rules be used verbatim. This is mostly proof-of-concept for where to write what we develop.

Please note that I am merging from my feature branch to the "Power Suit Superbranch" where I suggest we locate all work on power armor and merge into dev then to RC only once the monolithic power suit system is ready. The Power Suit Superbranch is where we will also make the required addendums to Specialist to appropriately differentiate them. Assassin and Priest will undergo similar changes in this branch to make their interactions with power suits explicit.

### DIFF
--- a/play/PlayerRules/CharacterCreation/Class-Specific Documentation/Power Suits/Chassis.txt
+++ b/play/PlayerRules/CharacterCreation/Class-Specific Documentation/Power Suits/Chassis.txt
@@ -1,0 +1,19 @@
+		Light Chassis
+Chassis Name:	Mantis
+Chassis Power Req:	3 PU
+Standard Speed:		1m/action
+Maneuver Bonus:		2m/action
+
+Fort Bonus:		2
+Reflex Penalty:		30
+Perception Penalty:	25
+
+Disrupt DC:		75
+Pwn DC:			120
+
+
+		Medium Chassis
+
+
+
+		Heavy Chassis
--- a/play/PlayerRules/CharacterCreation/Class-Specific Documentation/Power Suits/Power Suit Rules.txt
+++ b/play/PlayerRules/CharacterCreation/Class-Specific Documentation/Power Suits/Power Suit Rules.txt
@@ -1,0 +1,37 @@
+		Overview
+	Some militaries in the Terran Empire have employed the use of 
+large mechanical warsuits operated by a single pilot. This class of
+armor-vehicle hybrid is colloquially called "Power Suits." Power 
+Suits tend to be slow, tough and pack an ammount of punch rarely 
+seen by infantry. 
+
+		Power
+	As the name implies, most power suits carry a power plant
+on them that runs their various weapons, parts and chassis. A
+Power Suit may carry, but not activate or use a part or weapon
+that would exceed the power capacity of the suit's power plant.
+
+		Modes
+	Power Armor has three modes. Switching between modes takes
+Two actions for a pilot in the armor. These modes are:
+		- Combat
+		- Maneuver
+		- Shutdown
+	In combat mode, the suit works as expected. The pilot
+may direct the armor to move, fire and scan using its various 
+actions.
+	In Maneuver mode, the suit cannot use any of its active
+components or weapons. However it gains a bonus to move speed as
+listed in the chassis's Maneuver Bonus. 
+	In Shutdown mode, the suit is much more difficult to hack.
+However the suit cannot scan, move nor use active components or 
+weapons. If the Power Suit's power output no longer meets or 
+exceeds its power requirement, it immediately enters shutdown mode.
+
+		Cover
+	Power Suits are much larger than standard infantry, and 
+are typically a little less graceful to boot. They Cannot use the
+hunker action. Furthermore partial cover has no effect on the 
+accuracy of an enemy firing at a Power Suit and good cover acts 
+as parital cover for power suits. Due to their sheer size, enemies
+get a flat -20% to miss when firing at power suits. 
--- a/play/PlayerRules/CharacterCreation/Class-Specific Documentation/Power Suits/Parts.txt
+++ b/play/PlayerRules/CharacterCreation/Class-Specific Documentation/Power Suits/Parts.txt
@@ -1,0 +1,9 @@
+		Power Plants
+	Almost all Power Suits run on some sort of power supply. You'll
+need one in order to run your weapons, active components and to keep
+your chassis walking.
+
+Lion U-235 core		$1400
+Power Output:	8 PU
+Size:	1 slot
+


play/PlayerRules/CharacterCreation/Class-Specific Documentation/Power Suits/Chassis.txt
```diff
@@ -0,0 +1,19 @@
+		Light Chassis
```
1sourcecontrol, line 1: What the difference between a light chassis and a heavy one? I'd typically see light as the faster, less armored, less powerful version

play/PlayerRules/CharacterCreation/Class-Specific Documentation/Power Suits/Power Suit Rules.txt
```diff
@@ -0,0 +1,37 @@
...
+armor-vehicle hybrid is colloquially called "Power Suits." Power 
+Suits tend to be slow, tough and pack an ammount of punch rarely 
+seen by infantry. 
+
+		Power
+	As the name implies, most power suits carry a power plant
+on them that runs their various weapons, parts and chassis. A
+Power Suit may carry, but not activate or use a part or weapon
+that would exceed the power capacity of the suit's power plant.
+
```
1sourcecontrol, line 13: Sounds great

play/PlayerRules/CharacterCreation/Class-Specific Documentation/Power Suits/Power Suit Rules.txt
```diff
@@ -0,0 +1,37 @@
...
+	In combat mode, the suit works as expected. The pilot
+may direct the armor to move, fire and scan using its various 
+actions.
+	In Maneuver mode, the suit cannot use any of its active
+components or weapons. However it gains a bonus to move speed as
+listed in the chassis's Maneuver Bonus. 
+	In Shutdown mode, the suit is much more difficult to hack.
+However the suit cannot scan, move nor use active components or 
+weapons. If the Power Suit's power output no longer meets or 
+exceeds its power requirement, it immediately enters shutdown mode.
```
1sourcecontrol, line 29: If the suit is in shutdown, by this description, does this mean it's in a 'sleep' mode? Or is this an actual shut-off? If I shutdown my computer the software is realistically unhackable, as the communications don't have any power... vs "much more difficult"

play/PlayerRules/CharacterCreation/Class-Specific Documentation/Power Suits/Power Suit Rules.txt
```diff
@@ -0,0 +1,37 @@
...
+
+		Cover
+	Power Suits are much larger than standard infantry, and 
+are typically a little less graceful to boot. They Cannot use the
+hunker action. Furthermore partial cover has no effect on the 
+accuracy of an enemy firing at a Power Suit and good cover acts 
+as parital cover for power suits. Due to their sheer size, enemies
+get a flat -20% to miss when firing at power suits.
```
1sourcecontrol, line 37: Unless I misunderstand your wording... I don't think it's necessary to double stack penalties against the power suit: I'd give the suits -20% to miss against __OR__ the cover nerfs (and no hunkering in either case). As far as a I can tell In terms of realism, the -20% miss chance and cover nerfs are the same physical concept being penalized, large size = easier to hit so we can probably get away with 1 concept vs 2 to lessen the complexity. If that makes sense.

If you agree, it might be good to use the -20% miss chance. It's less intuitive, imo, but it allows the suit to be hit better in no cover, causes light cover be of no normal effect and heavy to be ~20% . 

By the way, can power armor crouch or go prone?

play/PlayerRules/CharacterCreation/Class-Specific Documentation/Power Suits/Chassis.txt
```diff
@@ -0,0 +1,19 @@
...
+Standard Speed:		1m/action
+Maneuver Bonus:		2m/action
+
+Fort Bonus:		2
+Reflex Penalty:		30
+Perception Penalty:	25
+
+Disrupt DC:		75
+Pwn DC:			120
+
```
1sourcecontrol, line 13: I suggest we use caution when giving fort bonuses, since that will grant additional death saving throws.

Why are you giving a perception penalty?

Also, where is the number of slots in the chassis stated?

1sourcecontrol (changes requested):
Some theory questions; mostly just a concern over powered armor and cover. Otherwise, looks great!

play/PlayerRules/CharacterCreation/Class-Specific Documentation/Power Suits/Chassis.txt
bleehu: /shrug?
No idea! Want to use "light" chassis to lend their mobility to the creation of disruption tanks, medium chassis maybe have more slots which will tend toward making damage tanks and heavy chassis easier to armor up to make an anchor tank?

or we could thow the class distinctions out of the window and have each chassis be unique unto themselves. I have no preference.

play/PlayerRules/CharacterCreation/Class-Specific Documentation/Power Suits/Chassis.txt
bleehu: We don't need fort bonuses specifically, but we do need some mathematical way to represent that force pushing or knocking over a power suit is harder than doing the same thing to an infantryman. Perhaps we should make a design doc for chassis, weapons and parts?

A perception penalty helps exacerbate the weakness to enemy assassins, marksmen and hunters. It also provides an avenue for social interplay. Having heavy power armor obscure your vision with bulletproof plates provides the opportunity to have more keen-eyed and rapidly-moving teammates spot targets and call in the support of the power suit.

We should probably add that, huh?

play/PlayerRules/CharacterCreation/Class-Specific Documentation/Power Suits/Power Suit Rules.txt
bleehu: I personally can't pull it off, but ya, your computer can actually be hacked when it's turned off. It's a very difficult side channel attack, but it can be done. long explanation for "it's really shut off."

play/PlayerRules/CharacterCreation/Class-Specific Documentation/Power Suits/Power Suit Rules.txt
bleehu: It think one of those double penalties will be negated by the defensive bonuses that we are going to give power suits. The remaining penalty helps to balance the cool abilities that we are giving power suits. I vote that we leave it, but I want this to be our project, not just mine. With that explanation, please let me know if I've swayed you or you still see it as a double negative. 

I think power armor should not be able to go prone at all. Maybe a crouch with some extended action cost?

play/PlayerRules/CharacterCreation/Class-Specific Documentation/Power Suits/Chassis.txt
bleehu: I don't intend to finalize the stats that we are tracking in each in this pass. Rather I hope to create separate branches for each atomic section of work after this.

play/PlayerRules/CharacterCreation/Class-Specific Documentation/Power Suits/Chassis.txt
1sourcecontrol: I think we should consider throwing the distinctions out of the window. Then we can name things like "ECSC Jaeger Assault Chassis". I'm sure we'll still use terms like "light, medium, heavy" for descriptors. Also opens the door wide open to unique sweet loot

bleehu: Lets make it so!

play/PlayerRules/CharacterCreation/Class-Specific Documentation/Power Suits/Power Suit Rules.txt
1sourcecontrol: Ok well, you can't do it wirelessly unless the system has a wirelessly powered chip. So, splitting hairs? Side channel attack meaning, applying power to the circuit with a soldering iron and some wires? I'm talking like, you plug in a usb cable and try to hack the computer. If the USB port is unpowered and the computer is off -no power flowing to ram, processor, bridges, I/O,- then I imagine you'd have to interface with the circuit board, which is impossible in a 4-6 second combat turn unless the assailant has the strength of the hulk , dexterity of a doctor, agility of a robot, and perfect knowledge of the system.

Or, are you talking about hacking it so that the ram has some crap in it that takes over the system on startup or would the player immediately lose control if successfully hacked?

play/PlayerRules/CharacterCreation/Class-Specific Documentation/Power Suits/Chassis.txt
1sourcecontrol: We could use fort and specify -does not grant additional death saving throws-, or go for the reverse and specify each piece exactly: +30  health, +6% Fort Save, +6% etc.

That implies a fairly low-tech suit then? But i'm with you on the concept.

👍

play/PlayerRules/CharacterCreation/Class-Specific Documentation/Power Suits/Power Suit Rules.txt
1sourcecontrol: So the net effect is:
no cover | partial | full
     -20%  | -20%  | 0%
We can do it your way. If we can, let's try to reword it  to make it more intuitive. 
Although if we have a squishy powered armor -the damage variant- then they will be easy targets?

Sure. We could extend the cost to 1/1 for activation/deactivation respectively, instead of 0/1.

1sourcecontrol (approved):
>I don't intend to finalize the stats that we are tracking in each in this pass. Rather I hope to create separate branches for each atomic section of work after this.

In the spirit of this last, I say shipit!

One of us should add things about crouching into the rules section when we get a chance

Issue #47, opened 2017-04-11
max stats at lvl one not listed


Wildlovelies:
more for a note on min maxing, but no where is it stated that 10 is the max you can put non bonus points into so i could dump stat charisma, luck and int and run a 20 strength berserker with a power fist

trowl223:
This is true. Intended rules: Max you can put into any one stat (even after cyborg stat increases) at level one is 10. The maximum for any stat is 15.

bleehu:
Don't write it here! Either put it in the rules or direct someone on the team to write the fix for ya!

bleehu:
I see this fix as a sort of band-aid as opposed to a well-thought-out fix. We will then have the same issues as soon as a character is level 2 that we have now. It also doesn't give the chance to grow more once at higher levels.

bleehu:
Pull Request submitted. Issue will close on acceptance.

bleehu:
Merged. Closing

PR #117, opened 2017-07-03
put a limit on player stats.

bleehu:
Intended to solve Issue #47 and prevent the abuse @wildlovelies pointed out with Cyborg especially.

Can I get a Shipit?

### DIFF
--- a/play/PlayerRules/Basic Rules.txt
+++ b/play/PlayerRules/Basic Rules.txt
@@ -200,10 +200,10 @@ Assigning Stats to a new character
 This will tell you a lot about your character. When assigning your stats,
 no stat can be higher than a 9 before applying racial/species bonuses and
 no stat can be lower than a 1 after applying racial/species bonuses. 
-Additionally, even after race/species bonuses, feats and equipment, no stat
-can ever be over 15.  The number of total stat points that you have to apply 
-should be listed with your Race or Species. See chapter 1: Races if you 
-haven't yet. 
+Additionally, even after race/species bonuses, feats, equipment and base stat 
+bonuses from leveling, no stat can ever be over 15.  The number of total base 
+stat points that you have to apply is listed with your Race or Species. See 
+chapter 1: Races if you haven't yet. 
 
 Leveling up:
 Gain feats according to the table below.
--- a/play/PlayerRules/Basic Rules.txt
+++ b/play/PlayerRules/Basic Rules.txt
@@ -201,9 +201,10 @@ This will tell you a lot about your character. When assigning your stats,
 no stat can be higher than a 9 before applying racial/species bonuses and
 no stat can be lower than a 1 after applying racial/species bonuses. 
 Additionally, even after race/species bonuses, feats, equipment and base stat 
-bonuses from leveling, no stat can ever be over 15.  The number of total base 
-stat points that you have to apply is listed with your Race or Species. See 
-chapter 1: Races if you haven't yet. 
+bonuses from leveling, no stat can ever be over 15 until your character is 
+level 15. After level 15, the limit for stat points is 20. The number of 
+initial base stat points that you have to apply is listed with your Race or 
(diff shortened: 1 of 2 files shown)


play/PlayerRules/Basic Rules.txt
```diff
@@ -195,6 +195,15 @@ Luck: Some people are just lucky! A GM will often have you roll luck to see
 	a coin off the ground right when an assassin is trying to shoot you in
 	the head. Luck also has an effect on what loot you find in dungeons.
 
+Assigning Stats to a new character
+	When making a new character, you'll have to assign their Base stats.
+This will tell you a lot about your character. When assigning your stats,
+no stat can be higher than a 9 before applying racial/species bonuses and
+no stat can be lower than a 1 after applying racial/species bonuses. 
+Additionally, even after race/species bonuses, feats and equipment, no stat
```
1sourcecontrol: > even after race/species bonuses, feats and equipment,

Probably should add something to signify the +4 attribute points gained by player leveling

play/PlayerRules/Basic Rules.txt
```diff
@@ -195,6 +195,15 @@ Luck: Some people are just lucky! A GM will often have you roll luck to see
...
 	the head. Luck also has an effect on what loot you find in dungeons.
 
+Assigning Stats to a new character
+	When making a new character, you'll have to assign their Base stats.
+This will tell you a lot about your character. When assigning your stats,
+no stat can be higher than a 9 before applying racial/species bonuses and
+no stat can be lower than a 1 after applying racial/species bonuses. 
+Additionally, even after race/species bonuses, feats and equipment, no stat
+can ever be over 15.  The number of total stat points that you have to apply 
+should be listed with your Race or Species. See chapter 1: Races if you
```
1sourcecontrol: I think it'd be better to write "The number of _Base_ stat points that you have to apply 
 _is_ listed with your Race or Species"
So that the manual sounds more assertive

1sourcecontrol (changes requested):

bleehu:
review changes made. Can I get a re-review and maybe a shipit?

play/PlayerRules/Basic Rules.txt
```diff
@@ -195,6 +195,15 @@ Luck: Some people are just lucky! A GM will often have you roll luck to see
...
 	the head. Luck also has an effect on what loot you find in dungeons.
 
+Assigning Stats to a new character
+	When making a new character, you'll have to assign their Base stats.
+This will tell you a lot about your character. When assigning your stats,
+no stat can be higher than a 9 before applying racial/species bonuses and
+no stat can be lower than a 1 after applying racial/species bonuses. 
+Additionally, even after race/species bonuses, feats, equipment and base stat 
+bonuses from leveling, no stat can ever be over 15.  The number of total base 
+stat points that you have to apply is listed with your Race or Species. See
```
1sourcecontrol: > "...and base stat bonuses from leveling"..

I think the player can get confused if you call it __base__ stat bonuses because:

>"The number of total base stat points that you have to apply"

We don't say what their total __maximum__  stat points is, only the starting stat points.

1sourcecontrol (approved):
Thoughts on this last review, then I'm fine with shipping it

play/PlayerRules/Basic Rules.txt
bleehu: Is that not clear with "no stat can ever be over 15" ?

1sourcecontrol: I'm talking specifically about the "total base stat points". When you say "total" it makes me think "maximum". But I know that you don't mean maximum-possible, you simply mean the maximum allotted by player race at the start of the game. 
However, I want to make sure players know what you mean, as well. Does that make more sense?

bleehu: Ah! that makes it much more clear! I'll make those changes and request another review.

Issue #103, opened 2017-05-19
Feature Request: Drunk Status Effect


bleehu:
We should have a condition for being intoxicated.

trowl223:
Yes. There should be two effects with various stages of intoxication : Uppers and Downers. They should be part of the "Sickened/Ill" condition set so a GM can add extra side effects for certain substances.

bleehu:
Does anyone else want it?

bleehu:
I'm taking it!

bleehu:
Awaiting @trowl223's ok for the alcohol warning.

trowl223:
Yeah. I don't think the game needs to be family friendly. Its something we can always take out of the game too.

bleehu:
We can't close the issue until the PR has been moved to dev (we probably shouldn't even close it until it's at least in RC, but we haven't been holding to that)

PR #121, opened 2017-07-07
Mh issue 103 drunk status

bleehu:
Maybe we'll add 'hung over' as a condition at a later date. 

We should note that by including this condition in the game, we would need to put a warning on the game to avoid marketing it to minors. I don't have a problem with that, but we need to be conscientious about what we're saying by including these effects.

Looking for feedback, especially a shipit/don't shipit.

### DIFF
--- a/play/PlayerRules/Basic Rules.txt
+++ b/play/PlayerRules/Basic Rules.txt
@@ -573,7 +573,7 @@ Hammered (drunk) - While hammered, you take a -30 to accuracy rolls when
 	heavy machinery. You take a temporary -2 to dex, perception and charisma.
 	If your level of intoxication increases while you are hammered, you black
 	out. Roll an additional fort check with DC equal to the intoxication increase
-	DC of the last drink you had. If you fail it, you lose 2 death saving throws 
+	DC of the last drink you had. If you fail it, you roll 1 death saving throw 
 	to alchohol poisoning. If you have no more death saving throws left, you 
 	are dead. If you succeed, lose no saving throws. You are now Downed.
 	As long as you don't black out, being hammered will turn into being 
--- a/play/PlayerRules/Basic Rules.txt
+++ b/play/PlayerRules/Basic Rules.txt
@@ -572,10 +572,7 @@ Hammered (drunk) - While hammered, you take a -30 to accuracy rolls when
 	attacking and a -50 to skill checks involving driving, flying or opperating
 	heavy machinery. You take a temporary -2 to dex, perception and charisma.
 	If your level of intoxication increases while you are hammered, you black
-	out. Roll an additional fort check with DC equal to the intoxication increase
-	DC of the last drink you had. If you fail it, you roll 1 death saving throw 
-	to alchohol poisoning. If you have no more death saving throws left, you 
-	are dead. If you succeed, lose no saving throws. You are now Downed.
+	out and you are now Downed.
 	As long as you don't black out, being hammered will turn into being 
 	buzzed after two hours.
 
--- a/play/PlayerRules/Basic Rules.txt
+++ b/play/PlayerRules/Basic Rules.txt
@@ -559,7 +559,27 @@ reduced by 1, permanently (repairable, DC 60% Repair check), not stackable.
 Poisoned - d10 damage per turn. After 5 turns, some poisons may induce 
 unconsciousness. DC 50, Fortitude. Greater / Lesser poison increases damage 
 (+5/-5) and DC (+25,-10).
+Buzzed (drunk) - Many drinks in the galaxy contain a substance called alchohol.
+	When consuming these drinks, the effect should list a FORT save to avoid
+	increasing your level of intoxication. The first level of intoxication is Buzzed.
+	While buzzed, you gain a +10% to miss with any attack and a temporary -1
+	to Perception. At the same time, you get a +1 to charisma and luck. You also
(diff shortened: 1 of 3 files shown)


bleehu:
To address Issue #103

play/PlayerRules/Basic Rules.txt
```diff
@@ -559,7 +559,27 @@ reduced by 1, permanently (repairable, DC 60% Repair check), not stackable.
...
+	take a -20 to skill checks involving driving, flying or opperating heavy 
+	machinery. If you fail a Fort save to prevent increasing your intoxication
+	level while buzzed, you become hammered. As long as you don't become 
+	hammered, being buzzed will wear off in two hours.
+Hammered (drunk) - While hammered, you take a -30 to accuracy rolls when
+	attacking and a -50 to skill checks involving driving, flying or opperating
+	heavy machinery. You take a temporary -2 to dex, perception and charisma.
+	If your level of intoxication increases while you are hammered, you black
+	out. Roll an additional fort check with DC equal to the intoxication increase
+	DC of the last drink you had. If you fail it, you lose 2 death saving throws
```
1sourcecontrol: I don't think the potential to die should be nearly so high. For comparison, Bleeding out on the ground loses 1 death saving throw per turn. To drink so much alcohol that you lose 2 saving throws at once, would almost be suicidal. 
That's my reasoning, anywhos.
So, I recommend either having two easy-to-succeed death saving throws, or just losing one death saving throw.

play/PlayerRules/CharacterCreation/03_Feats.txt
```diff
@@ -208,6 +208,15 @@ Prerequisites: None
...
 +10% Crit chance, +5% Extra-Crit chance.
 
+Punch Drunk:
+Prerequisite: Fort > 3
+Take a +5% to miss on all attacks when not drunk. 
+Instead of taking a penalty to accuracy when using melee attacks when 
+intoxicated, you get a -5% to miss when buzzed and a -15% to miss when
+hammered. These bonuses only apply to melee attacks. You also get a 
++10% on rolls to avoid blacking out.  
+
```
1sourcecontrol: I think this should be buffed. It feels to me like you are in net equilibrium after this feat, ie., not gaining anything (especially since 'getting hammered' offers across the board maluses)
I recommend changing this:
>Take a +5% to miss on all attacks when not drunk. 

to "Take a +5% to miss on all _ranged_ attacks when not drunk."
or "Take a +5% to miss on all _melee_ attacks when not drunk."

1sourcecontrol (changes requested):
See notes, below.

Also, what kind of warning would we have to put on the box? Surely minors can still play the game, even with alcoholic content, right? I can watch a PG-13 movie that has alcohol in it, and League of Legends has a T rating with tobacco and alcohol in it.

If it's just a warning that "we don't condone minors using substances", or something of that sort, then i'm fine with it.

bleehu:
There's a parental warning equivalent of rating our game/book T or PG-13 for alcohol use/abuse. I'm ok with it too, but I'd like @trowl223's express permission to add this before we do since this is going to affect his creative vision for how the game is played and who our target audience is.

play/PlayerRules/CharacterCreation/03_Feats.txt
bleehu: This partly reflects a personal belief of mine. I think getting hammered is _always_ a bad tactical decision. But the feat lets a person play a Haymitch(Hunger Games) or a practitioner of Zui Quan (drunken boxing) character that tries to stay buzzed at all times but has to not be hammered to be effective.

1sourcecontrol: That sounds fine to me; my complaint is on the mechanical level. Taking this feat kind of forces you to play at a disadvantage, and that seems self defeating. The feat effectively gives you a -1 to perception for ranged fighting while not drunk, and being drunk already gives you a -1 perception. And your stated target was for the relevant player to be buzzed, which only yields a +5% melee accuracy while buzzed with the feat. Oh, and I'm not sure if this is intentional, but the feat only states that you no longer have maluses to melee attacking with the buzzed/hammered, but it doesn't say anything about ranged attacks.

1sourcecontrol: @bleehu thoughts?

trowl223:
Permission granted, as mentioned in #103. Pending approval based on the downed / Death Saving throw clause . . . .

play/PlayerRules/CharacterCreation/03_Feats.txt
trowl223: Feats should grant +20% buffs, my rec. is to balance to that.

bleehu:
I'll yank the death saving throws, but not having a measured way to bleed out is going to make a Marvel Universe cockroach effect which is a very serious problem. Player characters are already dangerously too survivable to as it is. I'll open an issue to continue the discussion.

bleehu:
Changes made, Requesting one final review before shipit.

1sourcecontrol:
@bleehu, for some reason it looks like you aren't getting notified of the comments to the review. I asked for your thoughts 8 days ago. Maybe it's because the comment is under "show outdated"? 

*EDIT* my mistake, I see that you have made some actions regarding the conversation, you just didn't reply to the thread

1sourcecontrol (approved):

PR #150, opened 2017-09-01
2.0.4 r c to play

bleehu:
There's no point in doing all of this work if we don't publish it so that we can actually _play_ with it!


bleehu:
It says there are conflicting files, but there's no discernible conflicts?

bleehu:
I have no idea  why these were considered merge conflicts. But there we go.

trowl223 (approved):
Looks good!

PR #151, opened 2017-09-01
added a feat for getting better at swimming.

bleehu:
When new players were asked to pick feats, they complained that certain classes had a bunch of feats, but some like doctor, hunter, etc didn't. So I figured I'd make a number of small atomic additions to the feats table to help round it out.

This feat in particular is really applicable to anyone though. Which I think is actually a good thing.

### DIFF
--- a/play/PlayerRules/CharacterCreation/03_Feats.txt
+++ b/play/PlayerRules/CharacterCreation/03_Feats.txt
@@ -139,6 +139,8 @@ were on the college swim team, or maybe your military training had an
 emphasis on endurance and core strength developed in the pool. In any case,
 you gain 10 skill in swimming and a bonus +15 to checks when holding your
 breath. You also ignore movement penalties from wading through deep liquids.
+These bonuses increase in an additional 10 skill in swimming and an additional
++5 bonus to holding one's breath after level 7.
 
 First Shot Counts: 
 Prerequisites:
--- a/play/PlayerRules/CharacterCreation/03_Feats.txt
+++ b/play/PlayerRules/CharacterCreation/03_Feats.txt
@@ -133,7 +133,7 @@ basic computer inquiry or operation. +10% to ONE pertinent class skill
 this feat.
 
 Fish To Water
-Prerequisites:
+Prerequisites: Fort > 2
 Maybe you took a lot of swimming lessons when you were little, maybe you
 were on the college swim team, or maybe your military training had an 
 emphasis on endurance and core strength developed in the pool. In any case,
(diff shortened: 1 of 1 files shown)


play/PlayerRules/CharacterCreation/03_Feats.txt
```diff
@@ -132,8 +132,18 @@ basic computer inquiry or operation. +10% to ONE pertinent class skill
 (hacking, etc.) while using this feat, to be chosen at the time of taking 
 this feat.
 
-First Shot Counts: 
+Fish To Water
 Prerequisites:
```
1sourcecontrol: can we add a prereq like... FORT 3? It probably makes more sense to have dex, but i don't think dex needs to be any more valuable

play/PlayerRules/CharacterCreation/03_Feats.txt
```diff
@@ -132,8 +132,18 @@ basic computer inquiry or operation. +10% to ONE pertinent class skill
...
-First Shot Counts: 
+Fish To Water
 Prerequisites:
+Maybe you took a lot of swimming lessons when you were little, maybe you
+were on the college swim team, or maybe your military training had an 
+emphasis on endurance and core strength developed in the pool. In any case,
+you gain 10 skill in swimming and a bonus +15 to checks when holding your
+breath. You also ignore movement penalties from wading through deep liquids.
+These bonuses increase in an additional 10 skill in swimming and an additional
++5 bonus to holding one's breath after level 7.
```
1sourcecontrol: Can you change the wording to "These bonuses increase to 20 skill in swimming and a +20 bonus to holding one's breath after level 7."

Keeping the player from having to do math will be beneficial :)
I really like this leveling feature!

1sourcecontrol (changes requested):
Left two comments; seems pretty good!

My biggest concern is that players have a high potential of taking this feat in ignorance with a campaign that won't ever use water. Up to this point, In the campaigns I've played in, I've seen little water. And no reason to get into the water, to boot.

I see two easy solutions: 
1. Make "Fish To Water" a feat that is a prereq/one of the prereqs for a greater athlete/expert athlete feat.
2. Remake "Fish To Water" so that it incorporates more commonly used athletic skills

play/PlayerRules/CharacterCreation/03_Feats.txt
1sourcecontrol: I want to add a prereq so that we can favor less powerful primary stats and/or make a min-maxing player's decisions more impactful

play/PlayerRules/CharacterCreation/03_Feats.txt
bleehu: I really like that wording better. Will change! :)

bleehu:
I kinda don't want to merge this without giving @trowl223 a chance to get final say on it.

bleehu:
After talking with 1source and Trowl, both strongly suggested adding a prereq. Will add.

play/PlayerRules/CharacterCreation/03_Feats.txt
trowl223: Yeah, maybe a 3 in Fort.

PR #152, opened 2017-09-04
added class primary stats to classes.

bleehu:
Strength - 4
Dex - 4
Int - 4
Per - 3
Fort - 1
Charisma - 0
Luck - 0

These class primary stats will be the determining factor of how many nanites you get.


1sourcecontrol (approved):
Gucci

PR #155, opened 2017-09-25
MH-Basic-Rules: Dying details

bleehu:
As per Issue #133 
Also took the opportunity to explicitly list the DC for stabilization.

PR #156, opened 2017-09-25
Mh death crawl feat

bleehu:
I don't see this as being a very popular feat, since nobody plans to get downed, but it could be interesting nonetheless.

### DIFF
--- a/Basic Rules.txt
+++ b/Basic Rules.txt
@@ -352,11 +352,38 @@ Medic-kit, medic user: Works as normal, but add 30 to amount healed.
 
 
 ==============================
-Death
+Death, Dying and Downed
 ==============================
 Oh no! You have less than 1 health point left! What now!?!
 
 The good news is you are not* dead yet. You are bleeding out. 
+When your character falls below 1 hit point, you get 3 saving throws called Death 
+Saving Throws. These throws are not tied to any attribute. When you fail any three 
+times, you die. If you succeed, you do not lose one of these three Death 
+Saving Throws, but will still have to roll next round if you are not stabalized. 
+Once you have expended a saving throw, the next time you fall 
+below 1 health point your saving throws fall to 2. the next time, you have 1. you 
+can only replenish these saving throws by going a whole encounter without having 
+to use any. Each encounter that you go without expending a saving throw 
+replenishes one saving throw up to a maximum of three.
+
+Downed
+The state of having less than one Health Point is called being "downed." It is also
+sometimes called "bleeding out." They are equivalent. While a character is downed,
+they cannot move, cannot stand under their own power and fall prone if not 
+supported by other means. They cannot speak, but can mumble incomprehensively.
+Every turn that a character takes while they are downed, they must make a Death
+Saving Throw of DC 50. Success and failure of these throws is detailed above.
+
+Stabalize
+As an action, a teammate may bring you back up to 1 health point by suceeding 
+on a Medicine skill check to stabalize you. Once stabalized, you may take actions 
+and speak as normal. Don't forget that a recently stabalized character may need 
+to take an action to retrieve a dropped weapon and/or take an action to stand 
+back up if you have fallen.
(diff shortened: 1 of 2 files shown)


1sourcecontrol:
Prone movement happens to be the same speed limitations as this ability. I suggest lowering the speed cap to 3. 

In light of the fact that a player that goes down is more likely than ever to die (with the up and coming changes to death saves), and the fact that this feat would be more rarely used, I think the Nanite cost should be removed. To still have limits, you could put in a Max of 2 feat uses per instance of being downed, or just 1 use of the feat per turn.

If my above argument isn't compelling, please consider the following: if the player has no nanites left in their pool and goes down, then the feat does nothing. So, then, a rarely used feat has a high cost and moderate potential to have no effect at all (especially if you are a mage getting picked off in the middle of a fight). Also, the feat has a payoff that is only beneficial if you have teammates who will help you/be able to help and if you can reach them before you die. I'd consider that feat to be underpowered.

1sourcecontrol:
(I hope that didn't come across as too critical. I think the idea of this ability is great.)

bleehu:
I think it's reasonable to leave the crawl speed the same? I see your point about removing the nanite cost.

1sourcecontrol:
I think it's reasonable to leave it the same. My comment is mostly from a realism perspective.

Basic Rules.txt
```diff
@@ -350,6 +350,47 @@ Using a medikit:
...
+The good news is you are not* dead yet. You are bleeding out. 
+When your character falls below 1 hit point, you get 3 saving throws called Death 
+Saving Throws. These throws are not tied to any attribute. When you fail any three 
+times, you die. If you succeed, you do not lose one of these three Death 
+Saving Throws, but will still have to roll next round if you are not stabalized. 
+Once you have expended a saving throw, the next time you fall 
+below 1 health point your saving throws fall to 2. the next time, you have 1. you 
+can only replenish these saving throws by going a whole encounter without having 
+to use any. Each encounter that you go without expending a saving throw 
+replenishes one saving throw up to a maximum of three.
```
trowl223, line 368: An example of intended mechanic:
Player falls to -4 hit points. On their next turn, they roll a 23, failing the saving throw. Next turn, they roll a 17, bad luck. Before their next turn a comrade gets to them and uses a medical kit to patch them up. 4 turns later, the character walks on a landmine and is once again downed. Previously, they had 3 fail opportunities and used two. This time, they have 2 fail opportunities. On their next turn, they roll a 48, failing the throw. On the turn after, they roll an 89 and are safe for the round. Another buddy gets to them and stabilizes them. The encounter ends.

Later that day, the group is ambushed, and our unfortunate player is ganked. On their next turn, they roll a 92 and are safe. But before a friend gets to them in time, they roll a 10 on their next turn. They are now dead.

Lets rewind to before the ambush. Our example player decides instead of leading the group into the dark mine, they are going to provide rear cover. In the ambush, they are not downed, and after the encounter ends, they are still standing. (They have now gone one full encounter without being downed). The next day, a brawl breaks out in the restaraunt, one thing leads to another, and the player is enveloped by a plasma grenade which reduces their hit points to < 0. They have regained one Death Saving Throw after the ambush from the previous day. On their first turn, they roll a 2, failing the DC. The turn after that, they roll a 37, also failing the DC. They are now dead.

1sourcecontrol: Sounds great! :)

bleehu: Makes sense to me. This change is supposed to be tracked in the death and dying rules change in PR #155 though. This PR is specifically about the added feat.

1sourcecontrol: lolol I didn't notice that we were on a different PR, woops

1sourcecontrol (approved):
I approve of the feat as is (since it isn't broken), but I think it should be buffed. I know we made DSTs a raw 50/50 but for the cost of a feat I'd think it's very reasonable to give the player at least a +5% on DSTs as well as the movement.

PR #157, opened 2017-09-25
MH last stand feat

bleehu:
Please note that it would be a bad mistake to add the last stand feat without also adding rules as to what happens when an enemy retaliates to a player who is last-standing.

### DIFF
--- a/Basic Rules.txt
+++ b/Basic Rules.txt
@@ -352,11 +352,38 @@ Medic-kit, medic user: Works as normal, but add 30 to amount healed.
 
 
 ==============================
-Death
+Death, Dying and Downed
 ==============================
 Oh no! You have less than 1 health point left! What now!?!
 
 The good news is you are not* dead yet. You are bleeding out. 
+When your character falls below 1 hit point, you get 3 saving throws called Death 
+Saving Throws. These throws are not tied to any attribute. When you fail any three 
+times, you die. If you succeed, you do not lose one of these three Death 
+Saving Throws, but will still have to roll next round if you are not stabalized. 
+Once you have expended a saving throw, the next time you fall 
+below 1 health point your saving throws fall to 2. the next time, you have 1. you 
+can only replenish these saving throws by going a whole encounter without having 
+to use any. Each encounter that you go without expending a saving throw 
+replenishes one saving throw up to a maximum of three.
+
+Downed
+The state of having less than one Health Point is called being "downed." It is also
+sometimes called "bleeding out." They are equivalent. While a character is downed,
+they cannot move, cannot stand under their own power and fall prone if not 
+supported by other means. They cannot speak, but can mumble incomprehensively.
+Every turn that a character takes while they are downed, they must make a Death
+Saving Throw of DC 50. Success and failure of these throws is detailed above.
+
+Stabalize
+As an action, a teammate may bring you back up to 1 health point by suceeding 
+on a Medicine skill check to stabalize you. Once stabalized, you may take actions 
+and speak as normal. Don't forget that a recently stabalized character may need 
+to take an action to retrieve a dropped weapon and/or take an action to stand 
+back up if you have fallen.
+The skill DC to stabalize a fallen comrade is 80. Medicine skill, INT mod, and any
+bonuses from medical equipment such as medpacks may be added to that roll. 
+
 
 *Footnote: Some hazards can reduce you to zero health points and kill you 
 instantly. Some examples are falling several stories, lava baptisms,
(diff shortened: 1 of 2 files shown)


1sourcecontrol:
So if I'm reading this right:
A player needs to have three successful death saves and spend 120 nanites so that they can draw, aim, and fire a pistol or smg at 2 miss chance?

bleehu:
I think it's more likely that a character with this feat will go down with their akimbo weapon out. But ya. I don't want last stand to break the cost-benifit analysis of playing recklessly. And I think the irrational analysis of the "cool factor" will lead to players taking it despite its underpowerd-ness.

bleehu:
I should probably lower the nanite cost though.

1sourcecontrol:
I'll have to read the Downed rules again but don't players drop their weapons when they are Downed? 

Here's how I look at it: the cost of going down is huge, you have a very good chance of dying. This feat won't stop you from dying and only does something 50% of the time. If you removed the Nanite cost and gave them a free aim action as well as a firing action for a successful save, I would probably consider it a balanced feat

1sourcecontrol:
I'm not upset or anything, just want to make sure it's not underpowered

1sourcecontrol:
Do smgs get to be in full auto at 2 miss chance?
Also, should there be a range limit? Or can a marksman make a 500m glory shot while he's downed?

bleehu:
removed requirement to succeed on saving throw and lowered nanite cost.

1sourcecontrol:
Why do we need a Nanite cost? Is there a downside to letting a player automatically use this feat when they go down?

Also, i believe 40 Nanite cost is higher than any other non-mage ability in C2.0. with the average player having a 60 Nanite pool, this cost is enormous.

Also, what is the to-miss chance of 2? Does that mean that a player will always have a 2 miss chance when using this feat, regardless of range and their own abilities? Or does this mean their weapon now has a 2 miss chance, no matter the range, and their other perception bonuses apply?

1sourcecontrol:
(although the real-life average player will probably tend to Max their Nanite pool due to their primary stat; so I expect the real average to be around 80 or 90. That means this ability will consume about half of the average player's pool. Cost still seems incredibly high)

bleehu:
There is an incredible tactical advantage to being able to sit up after being knocked down which I'm trying to counterbalance. Maybe dropping it down to 30 or 25 would be appropriate, but I think being able to sit up without having to make the tactical plan to keep enough nanites to use when you do go down.
I'm trying to avoid using going down to last stand as a first plan strategy. Maybe last stand shouldn't be added at all?

1sourcecontrol:
What's the incredible tactical advantage?
I see a minor advantage, that on the player's following few turns they have the chance to do some damage.
A restriction should probably be added so that if they take damage after activating this ability, they are knocked out cold. Or something to that effect.

Are you talking about a FOO when you say 'first plan strategy'? 

I don't see this becoming a FOO strategy. If the first thing a player does in combat is run out and get knocked down, they're going to have to make a saving throw to use this ability, take up another player's turn to save them, and have less actions than a normal turn. Sounds like a very sub-optimal play style that will quickly lead to their own death or a death spiral of the whole team.

The ability only allows you to get more out of otherwise wasted turns. I don't see how a player can really abuse the game by planning on going down?

bleehu:
Maybe I should call it a "goto strategy". It's not just a FOO strategy that I'm concerned about; I'm trying to capture the player's tendency to play in less-than optimal ways. 

Maybe there's already enough mitigation, but this is what I'm worried about:
"as long as it's not instant-death damage, I can take it! 800+ damage, no problem! I'll just go down and last-stand! "

Coup-de-grace rule is being added as part of this PR: https://github.com/trowl223/Compound_X/blob/MH-Last-Stand-Feat/Basic%20Rules.txt#L377

Basic Rules.txt
```diff
@@ -350,6 +350,49 @@ Using a medikit:
...
+
+Downed
+The state of having less than one Health Point is called being "downed." It is also
+sometimes called "bleeding out." They are equivalent. While a character is downed,
+they cannot move, cannot stand under their own power and fall prone if not 
+supported by other means. They cannot speak, but can mumble incomprehensively.
+Every turn that a character takes while they are downed, they must make a Death
+Saving Throw of DC 50. Success and failure of these throws is detailed above.
+When downed, any additional attack that hits regardless of damage is the same as
+having failed one Death Saving Throw.
```
1sourcecontrol, line 378: >any additional attack that hits regardless of damage is the same as having failed one Death Saving Throw.

I think this will destroy gameplay. Any smg burst, any aoe damage from friend or foe, any pebble hits you, and *pop* all our efforts at balancing gameplay will be gone, and 3 or 2 death saving throws will hardly matter.

1sourcecontrol (changes requested):

1sourcecontrol:
Ah, thanks for the explanation. 
"I'm trying to capture " -- trying to what? Trying to stop it? Trying to make players do it more?

Besides the fact that going down is heavily penalized by losing death saving throws, we could just add a clause to being "Downed". Something to the effect: "If your health is reduced to -50, then you cannot use any feat or ability while you are down"

bleehu:
by capture, I mostly meant "communicate".

I think the -50 health is a really good fix. I think with that fix, we could get rid of Coup De Grace and still have this feat be balanced. Though I'm a little concerned that this feat could lead to some unexpected combos in the future. @trowl223 what do you think?

1sourcecontrol:
We can add it to the design doc as something to pay attention to when designing other abilities.

I think the Nanite cost should be removed, especially because the feat will be worthless if a player gets gunned down with an auto rifle and takes 300 damage. In fact, it may be very rare that players will go down with more than -50 HP. Also, I feel that Nanite cost is overly penalizing because it's already very dangerous to go down.

Or instead of Nanite cost removal, some other large buff to the feat. I'm speaking these things from the perspective of balancing feats to be of equal value.

You haven't answered my question about the miss chance defined in this feat.

PR #162, opened 2017-10-25
Combat2.0superbranch alex l merge to DEV

1sourcecontrol:
Pls review


bleehu (approved):
SHIPIT!!!!!! PARTYYY!!!!!!

play/PlayerRules/Basic Rules.txt
```diff
@@ -236,8 +330,28 @@ Luck:
...
+Movement Speed:
+	Your speed in meters, for every two actions. Use (Movement Speed)/2 for one action (drop the remainder). 
+	For more information, see topic "Movement and Sprinting", below.
+Carry Ability:
+	Not 'carrying' in the traditional sense. Carry ability cancels out movement speed penalties.
+	If you have 6 Carry Ability and 6m of movement speed penalties, you end up with no speed penalty.
+Skill Point Gain:
+	How many skill points you gain every level. Every time you level up, you gain skill
+	points according to your "Skill Point Gain". Note about gaining skill points: you have a 
+	max of 50 points in any one Skill. Also, you have a max of 17+3*[CHA] points in any one Charisma Skill.
```
bleehu: I really like having a max skill points limit. We needed that bad.

play/PlayerRules/Basic Rules.txt
```diff
@@ -351,8 +466,22 @@ Medic-kit, non-medic user: works as if you are a medic with skill 30.
...
+Short Rest:
+Taking about an hour, characters can take a Short Rest to rest and recoup which
+will restore all the character's Nanites.
+
+Long Rest:
+Taking about eight hours, characters need to rest after a long day of adventuring.
+If you choose to stay up all night, you won't be at your best for the next day.
+Restores all of a character's Nanites.
+
+==============================
```
bleehu, line 481: I like the ability to rest! We'll add more here later.

play/PlayerRules/Basic Rules.txt
```diff
@@ -396,8 +525,8 @@ other mercantile area. Ask the GM if there are weapon shops or kiosks around,
 they should point you to one. From there you can apply upgrades and 
 attachments.
 The GM may tell you specific attachments that are available, or they may say 
-you can just go about your usual customization. Only one attachment can be 
-used per section (unless specifically stated in the item''s description). 
+you can just go about your usual customization. See the Attachments document
+for more details on the kinds of attachments you can use, and their restrictions.
```
bleehu, line 529: Good reference, rather than re-stating

play/PlayerRules/Basic Rules.txt
```diff
@@ -407,192 +536,396 @@ Combat starts when either you or an enemy decides they would like to make an
...
+"Armor and Getting Hit" and "Damage Types and Armor Types"). Your damage dealt is often
+based on a combination of your DEX and STR. If you are wielding a weapon against
+an unarmed opponent, they may still incur damage depending on their skill.
+	NOTE: if you are fighting unarmed, you can (normally) deal full non-lethal damage or 
+half lethal damage (your choice). Non-lethal damage should be recorded as a separate (duplicate) 
+health bar. When their non-lethal health bar reaches 0, the target becomes unconscious/incapacitated.
+	*Modifiers from specific weapons or unarmed[x] attacks are listed in the weapons document.
+	**To determine bonus modifiers from skill, for every 5 skill points in a relevant 
+	skill (such as "melee combat skill") take a bonus +0.5 when rolling against guard DC.
+	Drop any decimal remainder after adding all the modifiers.
```
bleehu, line 575: It's so nice to have melee rules I feel like I can actually understand enough to run. And they're all even written down!

play/PlayerRules/Basic Rules.txt
```diff
@@ -407,192 +536,396 @@ Combat starts when either you or an enemy decides they would like to make an
...
+an opponent during grappling, but if the strike fails your opponent may attempt 
+to gain control of the grapple. Skill in grappling is different than skill in 
+unarmed combat.
+
+Kicking/Shoving:
+Instead of dealing full damage, a character can try to kick/shove a human sized
+opponent to force movement by rolling for a normal unarmed/armed melee attack 
+with an additional -2 miss penalty against the opponent's Guard DC. If the attack 
+succeeds, it yields 1/2 the damage it normally would and pushes the target 
+1m for every 5 strength you have.
```
bleehu, line 596: I extra like this!

play/PlayerRules/Basic Rules.txt
```diff
@@ -407,192 +536,396 @@ Combat starts when either you or an enemy decides they would like to make an
...
+instead of the normal 1x).
+
+Shock:
+When players take more than 70 damage, they must roll a Shock save at a DC of 
+the damage taken. On fail they pass out. on success they stay conscious. When 
+players fail, they may roll each turn at a DC of 70 to wake up. They continue 
+to roll either for (15 - FORT) turns, or until they pass.
+
+Stealth Attacks:
+If you are stealthed (undetected by an opponent) when you initiate an attack action,
```
bleehu: "stealthed" should be "hidden"

play/PlayerRules/Basic Rules.txt
```diff
@@ -407,192 +536,396 @@ Combat starts when either you or an enemy decides they would like to make an
...
+If your weapon has a minimum distance (usually because of a high-powered
+scope), and you fire inside that distance, you are forced to hip fire.
+
+Blind Firing:
+Add 6 to a ranged weapon's miss chances. Blind firing allows you to shoot around
+cover without looking or putting yourself at risk. You can blind fire while hunkered.
+Your blind fire miss chance is affected by your base weapon miss chance(s), not affected by
+attachments unless otherwise noted, reduced by PER stat combat bonus (as usual), and reduced
+by LUCK stat combat bonus. It's recommended to calculate this separate miss chance and 
+write it down before combat.
```
bleehu: I like that everyone can blind-fire. Should we adjust the blindfire feat?

play/PlayerRules/CharacterCreation/03_Feats.txt
```diff
@@ -42,275 +42,265 @@ Acquired: . . . Who Needs Nano-Technologies?
...
-When reloading, if you Critically succeed, the first round out of that 
-magazine deals Extra-Crit damage.
+Prerequisites: LCK 6
+When reloading, if your reload roll is higher than an 80 (including any applicable skills) or
+if you roll inside your crit range, your first shot is an automatic critical + 10 bonus damage,
+for a total of 30 bonus damage and other possible critical effects. (Instead, if you have Double Tap, 
+you get +10 bonus damage for each Double Tap bullet).
+You have to reload a fresh magazine in order to trigger the bonus again. In other words, you cannot use
+First Shot Counts more times in one encounter than the number of fresh magazines you are carrying. If
+your weapon damages an area, First Shot Counts does not take effect.
```
bleehu, line 141: Very good balancing!

play/PlayerRules/CharacterCreation/03_Feats.txt
```diff
@@ -42,275 +42,265 @@ Acquired: . . . Who Needs Nano-Technologies?
...
+	used as bats, fan blades as swords, etc., etc.
 
 Last Shot Counts: 
-Prerequisites:
-The last shot in every mag deals Extra-Crit damage, but your reload DC is 
-twice as high.
+Prerequisites: PER 5
+The last shot in every mag deals is an automatic critical + 10 bonus damage,
+for a total of 30 bonus damage. If your weapon has a magazine size of 1 or damages an area,
+Last Shot Counts does not take effect.
```
bleehu, line 175: Needs similar balances to First Shot Counts?

play/PlayerRules/CharacterCreation/03_Feats.txt
```diff
@@ -42,275 +42,265 @@ Acquired: . . . Who Needs Nano-Technologies?
...
-Run 'N' Gun: 
-Prerequisites: PER > 6, DEX > 6 or Soldier, Specialist classes.
-you may move at half speed while firing. +10% Miss Chance. If this feat is 
-performed while sprinting, half your movment speed before you add the 2m from 
-sprinting. The Miss Chance penalties stack. 
+NEEDS A BUFF:
+	Rainman: 
+	Prerequisites: CHA <3 and INT >8
+	Your character has something going for them. You may ask the GM how many 
+	bullets each opponent has left in their weapon.
```
bleehu, line 236: Dev note in code? Add an issue, not a line the rules!

play/PlayerRules/CharacterCreation/03_Feats.txt
```diff
@@ -457,76 +461,169 @@ Proficient: Explosives
...
+You may Roll to-hit your first target in order to determine if you hit all of these 
+targets. If you fail the roll, then you may roll again to hit your first target
+normally.
+
+Run 'N' Gun [10 Nanites]: 
+Prerequisites: PER 7 & DEX 7, or Soldier, or Specialist Classes.
+For 1 Action, you may move at half speed while firing. +1 Miss Chance. If you don't Aim before 
+using this feat, you must Hip Fire. If this feat is performed while sprinting, 
+half your movement speed before you add the +1m per-action from sprinting. The Miss
+Chance penalties stack for consecutive uses of Run 'N' Gun.
```
bleehu: I don't like run n gun being only for soldier or specialist. At all.

play/PlayerRules/CharacterCreation/03_Feats.txt
```diff
@@ -457,76 +461,169 @@ Proficient: Explosives
...
+		long rest. You can reroll as many times as you want (which uses up your
+		rerolls-per-rest) and if you reroll, choose which of the rolls to keep.
+		
+Positional Mastermind: 
+Prerequisites: Mutually Exclusive with Will to Survive; Charisma 9
+	You have incredible understanding of how beings in your universe perceive and react
+to their environments, allowing you to feint and position yourself in cover better. 
+If you have 9 Charisma, any living creature targeting you gets a +1 miss chance penalty
+when you have any kind of cover. If you have 12 Charisma, anything targeting you gets a 
++1 miss chance penalty when you have any kind of cover.
```
bleehu, line 599: no difference between 9 and 12 charisma?

bleehu (approved):
Still a few notes about issues to raise later, but nothing blocking the merge.

play/PlayerRules/CharacterCreation/03_Feats.txt
1sourcecontrol: note to me: all caps "living creature" and "anything"

play/PlayerRules/CharacterCreation/03_Feats.txt
1sourcecontrol: I think Taylor thought it was fine to allow _first shot counts_ to occur. I think at least one of these should be allowed, and it made more sense to allow first shot counts since it isn't necessarily an absolute that you will get bonus damage on your first shot. 
In other words, having both _last shot counts_ and _first shot counts_ to proc on single fire is a whole lot of damage, and in some ways, _first shot counts_ is the weaker feat (since it has a chance to fail) if we were going to allow one of them to occur.

play/PlayerRules/CharacterCreation/03_Feats.txt
1sourcecontrol: :(

play/PlayerRules/Basic Rules.txt
1sourcecontrol: :D

play/PlayerRules/Basic Rules.txt
```diff
@@ -236,8 +330,28 @@ Luck:
...
+	Some class feats and other abilities can increase the number of Nanites you can hold and use at a given time. 
+	Also, Unlike Health, Nanites regenerate during combat at a rate of 3 Nanites per turn. 
+
+==== Secondary Stat Details ====
+Movement Speed:
+	Your speed in meters, for every two actions. Use (Movement Speed)/2 for one action (drop the remainder). 
+	For more information, see topic "Movement and Sprinting", below.
+Carry Ability:
+	Not 'carrying' in the traditional sense. Carry ability cancels out movement speed penalties.
+	If you have 6 Carry Ability and 6m of movement speed penalties, you end up with no speed penalty.
```
Turtlelord26: If it doesn't have anything to do with carrying, can we call it something else for clarity? Maybe something along the lines of 'Stability'?

play/PlayerRules/Basic Rules.txt
Turtlelord26: While I agree with Mike that we did need some limiting mechanic in skills, I question a flat cap on the grounds that, say, an Engineer should always be able to be the clearly best repairman, not just as good as the party palawin. This is probably exactly the origin of the classic class vs cross-class skill divide, and I wonder if we might do something similar, with different caps and points to ranks ratios.
I will also note that since we have so little scaling in CX aside from gear, we could consider a dynamic skill cap based on level. It wouldn't have to grow very quickly, but it would help people feel that they're getting relevant stuff out of leveling up. Skills are the only thing that happen every level, and if someone maxes out all the skills that are particularly relevant to them by level 7 or 8, further skill gains are less exciting.

play/PlayerRules/Basic Rules.txt
```diff
@@ -407,192 +536,396 @@ Combat starts when either you or an enemy decides they would like to make an
...
+half lethal damage (your choice). Non-lethal damage should be recorded as a separate (duplicate) 
+health bar. When their non-lethal health bar reaches 0, the target becomes unconscious/incapacitated.
+	*Modifiers from specific weapons or unarmed[x] attacks are listed in the weapons document.
+	**To determine bonus modifiers from skill, for every 5 skill points in a relevant 
+	skill (such as "melee combat skill") take a bonus +0.5 when rolling against guard DC.
+	Drop any decimal remainder after adding all the modifiers.
+
+Hitting What You Intend:
+If you roll 3+ above your miss chance (3+ above the opponent's guard DC), you 
+ignore the opponent's armor and deal full damage.
```
Turtlelord26, line 579: 👍 👍 👍

play/PlayerRules/Basic Rules.txt
```diff
@@ -407,192 +536,396 @@ Combat starts when either you or an enemy decides they would like to make an
...
-During combat, most weapons are Semi-Auto, meaning that each time you pull the 
-trigger, the gun fires and chambers a new round automatically one time. It 
+
+==== Other Combat Details ====
+Critical Hits:
+Critical hits are accuracy rolls that land inside your critical hit range (aka. crit chance). 
+If you roll a critical hit, your weapon does +20 damage for that attack (and/or other effects,
+depending on your abilities). By default, everyone starts with a critical hit range of "1", 
+meaning, an attack roll of a 10 would be considered a critical hit. But your critical range may
+not be "1" because your Luck modifies your critical hit range: ((Luck-5)*0.4), drop the remainder,
```
Turtlelord26: Or, to save people a calculation, we could say (Luck Modifier / 10)?

Also, are you dropping the remainder or rounding down? This is relevant with negative luck modifiers. Notably, if you just drop remainders, there's very little difference between 3 Luck and 7, but if you round down the granularity is reduced to a range of three, from 2-4 or 5-7.

play/PlayerRules/Basic Rules.txt
```diff
@@ -407,192 +536,396 @@ Combat starts when either you or an enemy decides they would like to make an
...
+is added to your critical hit range. Which means, if you have a Luck stat of 8 (and thus, by
+the equation, +1 to your critical hit range), an attack roll of 9-10 would be considered a
+critical hit. 
+All weapons and melee can critical hit except for Fully-Automatic weapons (weapons that 
+shoot more than 1 bullet per Action by default).
```
Turtlelord26: This is a little off-handed. I hope full-auto inability to crit is stated somewhere else too, otherwise I expect this to be one of our more forgettable rules as-is. Could probably fix by making it its own sentence here, that would cause it to hold more emphasis.

play/PlayerRules/CharacterCreation/02_Classes.txt
```diff
@@ -2,645 +2,696 @@ Core classes:
...
- * to add the X skill to your skill  list, and it starts at rank 10, meaning a 
- * +10% to any roll for that skill.  At the first level taken in a class
- * (other than level 1), you get the  single star version of the feat. At 
- * level 5 and every 5 levels thereafter,  you can  choose to rank up the 
+ * For any listed class, if it says "Gain X Skill at 10" that means you get
+ * to add the X skill to your skill list, and it starts at rank 10, meaning a 
+ * +10% to any roll for that skill. At the first level taken in a class
+ * (other than level 1), you get the single star version of the feat. At 
+ * level 5 and every 5 levels thereafter, you can choose to rank up the 
  * feat, or add the other feat to your  character.
```
Turtlelord26, line 13: "you do not select a class feat until level 5"
"At the first level taken in a class // (other than level 1), you get the single star version of the feat"

Can we take this opportunity to annihilate the ambiguity?

play/PlayerRules/Basic Rules.txt
```diff
@@ -236,8 +330,28 @@ Luck:
 ╠═══════╬═══════════════════════════════════════════════════════════════════════════════════════╣
 ║ 9-10  ║ Nothing bad happens... this time							║
 ╚═══════╩═══════════════════════════════════════════════════════════════════════════════════════╝
+
+Nanites:
+	Although not a primary stat, The maximum number of Nanites you can use at a given time is determined
+	by your character class and an associated primary stat. Your class may tell you its "Class Primary Attribute" 
+	is Strength. In that case, Strength * 10 is your maximum Nanites. More generally:
+	    Your Maximum Nanites = (Your Class's Primary Attribute) * 10
```
Turtlelord26, line 338: Is this implementation replacing the proposed Stamina-Nanites split?

play/PlayerRules/CharacterCreation/02_Classes.txt
```diff
@@ -2,645 +2,696 @@ Core classes:
...
-Armour Proficiencies: Light, Medium, Heavy.
-Starting Wealth: $1700+100*Luck + 100
+=====   SOLDIER   =====
+You have the blood of a warrior flowing through you. Your sole 
+purpose is to fight until the threat is neutralized.
+Class Primary Attribute: Strength
+Passives: Gain Skill Weapon Use[X] at 10, Gain a +15 to Rolls for initiative. (You go first more often).
+Weapon Proficiencies: Pistols, Submachine Guns, Shotguns, Carbines, Assault Rifles, Battle Rifles.
+Armor Proficiencies: Light, Medium, Medium-Heavy.
+Starting Wealth: $850 + 100*Luck + 2920
```
Turtlelord26, line 23: Is anything keeping us from combining the 850s and the class-tailored numbers like 2920?

play/PlayerRules/CharacterCreation/02_Classes.txt
```diff
@@ -2,645 +2,696 @@ Core classes:
...
+	Significant past training or combat experience has given you honed battle reactions, 
+	adrenaline simply aids your ability to understand and react to the task at hand. 
+	+1 non-attacking action, this turn only. Activate as a free action 
+	during your turn, limit one use per turn and a max of 3 activations per encounter.
+	|
+	|
+	V
+	Maximum of 6 activations per encounter; costs [20 Nanites] to use.
+	----------------------
+	Battle Cry / Raise Moral -
```
Turtlelord26, line 36: Morale. By the way, I do greatly appreciate your efforts on behalf of grammar.

play/PlayerRules/CharacterCreation/02_Classes.txt
```diff
@@ -2,645 +2,696 @@ Core classes:
...
+	You roar as you lead the charge against the enemy. 1 Action. 50% + (CHA mod) chance to 
+	raise allied effectiveness 20% (or +2 if relevant) on each of their next rolls [including 
+	your next roll]. You must use your next action to run towards enemies [minimum 1m].
```
Turtlelord26, line 39: I understand the flavor here, but is it not also flavorful to scream Dramatically(TM) while making a Dramatic(TM) attack? Attacks like suppression, full-auto, or grenades?

play/PlayerRules/CharacterCreation/02_Classes.txt
```diff
@@ -2,645 +2,696 @@ Core classes:
...
-"Doctor" - You know how or why something works [most likely humans / humanoid 
-bodies), and put that knowledge to good use by supporting your team.
-Passives: Gain 5 skill in Anatomy, 5 skill in Pathology, and 10 skill in 
-medicine. Every level, you gain one skill in both of the "hippocratic" and "
-malpractice" skill trees. At level 1 you start with an arm mounted needle 
-launcher, handheld pathogen synthesizer, field dressing applicator and 
-a package of ten needles. You start each encounter with 200 adrenaline
-points and will need to keep track of adrenaline points, needles and 
+	----------------------
+	Needs More Dakka [10 Nanites] -
```
Turtlelord26, line 73: Seems cheap, at least for having no rate limit.

play/PlayerRules/CharacterCreation/02_Classes.txt
```diff
@@ -2,645 +2,696 @@ Core classes:
...
+	|
+	|
+	V
+	Your ferocity knows no bounds. Targets fired upon with Needs More Dakka
+	must make a DC 60 + (Charisma * 8) Shock check. On failure, target is 
+	Pinned Down (see section "Conditions" in the basic rules document).
+	----------------------
+	Heavier Ordinance Not Required [25 Nanites] -
+	Your expertise with heavy weapons is sufficient for the task at hand. 
+	Gain -2 Miss Chance and a bonus +25 damage for an attack with a heavy weapon.
```
Turtlelord26, line 89: "A" heavy weapon? Does that mean one in particular, one class of heavy weapon, or all heavy weapons?

play/PlayerRules/CharacterCreation/02_Classes.txt
```diff
@@ -2,645 +2,696 @@ Core classes:
...
+	You know that every breath you take could be your last, and one errant twitch
+	could mean the difference between sending your brother in arms to see his 
+	friends and family or his ancestors. You love it. Your maximum adrenaline is 
+	600 and you start each encounter with 240 adrenaline points instead of 200.
+	|
+	|
+	V
+	Adrenaline Addict - 
+	"Nothing in life is so exhilarating as to be shot at without result" ~ Winston Churchill
+	Any time an enemy shoots at you and misses, gain 10 Adrenaline Points.
```
Turtlelord26, line 125: I haven't yet seen adrenaline expenditures, but based solely on starting with 240, gaining 10 per miss seems quite anemic.

play/PlayerRules/CharacterCreation/02_Classes.txt
```diff
@@ -2,645 +2,696 @@ Core classes:
...
+	friends and family or his ancestors. You love it. Your maximum adrenaline is 
+	600 and you start each encounter with 240 adrenaline points instead of 200.
+	|
+	|
+	V
+	Adrenaline Addict - 
+	"Nothing in life is so exhilarating as to be shot at without result" ~ Winston Churchill
+	Any time an enemy shoots at you and misses, gain 10 Adrenaline Points. 
+	-----------------------
+	The Harmisist -
```
Turtlelord26, line 127: I really am trying to skip past purely grammatical matters, but it is important that 'The Harmacist' be spelled correctly.

play/PlayerRules/CharacterCreation/02_Classes.txt
```diff
@@ -2,645 +2,696 @@ Core classes:
...
+	Any time you build a construction, you may spend 15 Nanites to give that 
+	construction 50 additional health points.
+======================================
+
+
+=====   PRIEST   =====
+You've lived your life in the deepest recesses of the holonet, and have come out 
+alive and strong. You can figure out how a security system works, hack into systems,
+or even hack your allies' Nanites to perform better.
+Class Primary Attribute: Intelligence
```
Turtlelord26: I'm not necessarily asking for a change here, just making a note. If Medic switches to Int as I proposed above, I recognize that that might create something of an Intelligence bloat in the bigger picture. So, I would note here that I could see moving priests to Perception as a primary attribute while keeping Int in their spell descriptions as a efficacy gauge, in the same style quality/quantity argument I used in the Medic's case. After all, you do have to see the weakness before you can hack at it.

play/PlayerRules/CharacterCreation/02_Classes.txt
```diff
@@ -2,645 +2,696 @@ Core classes:
...
+Your specialized skills with technology are combined with battle 
 training, providing you an edge in combat situations.
-Passives: Gain 100 nanites. any number of these can be put into a permanent 
-passive CS Class B unit that is implanted in your body. When not wearing armor,
-it covers 75% excluding head, hands, and feet. when wearing armor, it extends 
-your DR by 10 and you gain 5% coverage.
-Weapon Proficiencies: Pistols, SMGs, Carbines and Battle Rifles.
-Armour Proficiencies: Concealed, Recon, Light, Medium.
-Starting Wealth: $1700+100*Luck - 400
+Class Primary Attribute: Strength
```
Turtlelord26: I'm really not sure about Adepts, but see the Str vs Fort argument I made for Soldier.

play/PlayerRules/CharacterCreation/02_Classes.txt
```diff
@@ -2,645 +2,696 @@ Core classes:
...
+=====   PALADIN   =====
+You are a righteous warrior, vetted and trained by one of the 
 powerful corporations or some other technologically advanced entity. You seek 
-to uphold the principals which define the group you''ve sworn allegience to. 
+to uphold the principals which define the group you've sworn allegiance to. 
 Through this devotion you have been granted access to high tech in order to 
 pursue your goals and missions.
-Passives: You gain 45 total nanites and each level you gain an additional 5 
-maximum nanites for each level after your first.
+Class Primary Attribute: Fortitude
```
Turtlelord26: I can see Fort, but I would have expected Cha, especially after moving Priests off of it.

play/PlayerRules/CharacterCreation/02_Classes.txt
```diff
@@ -2,645 +2,696 @@ Core classes:
...
 Tome of Spells, which must be on the Mage in order to cast spells. This Tome 
 is accessed as any typical holobook, and can be connected to Altars to expand 
 the Mage''s available spells.
 Weapon Proficiencies: Pistols.
-Armour Proficiencies: Recon.
-Starting Wealth: $1700+100*Luck - 1400
+Armor Proficiencies: Recon.
+Starting Wealth: $850 + 100*Luck + 1040
 Feats: The Mage picks one of the following branches at level 5. Instead of 
 gaining more class feats at levels 1-, 15 and 20, they gain access to more
```
Turtlelord26: 10 is typoed. In Sorcerer description too.

play/PlayerRules/CharacterCreation/02_Classes.txt
```diff
@@ -2,645 +2,696 @@ Core classes:
...
+If the Mage is the performer, the Sorcerer is the composer. They 
 shroud themselves in secrecy, operating at the highest levels of their 
 organizations. Typically they spend their time researching new uses of Nanites 
 and reverse engineering captured technology. As the designers and researchers, 
 Sorcerers generally don''t need to have a Tome of Spells handy to cast their 
 spells. However, this also means that their spell knowledge is limited to 
 their memory, so they frequently shift around a mental list of spells which 
 can be used.
-Passives: Knowledge Nanites - 10; +50 nanites, +10 per level.
+Class Primary Attribute: Intelligence
```
Turtlelord26: Not having wisdom in our system is unfortunate here. I propose that we do want to distinguish our primary casting classes with primary attribute, and since we don't have Wis to swap sorcerers to, what if we swapped mages to Cha? Unlike Sorcerers who are designing spells, mages get all their spell information literally laid out in front of them. As the description says, performers and composers.

play/PlayerRules/CharacterCreation/02_Classes.txt
```diff
@@ -2,645 +2,696 @@ Core classes:
...
+	Consumers are generally the product of a powerful Sorcerer falling into Evil 
+	ways. Their Technology favors negating other Nanite users'' abilities, 
+	stealing Nanites from others, and denaturing matter. Consumers are powerful 
+	direct combatants.
+======================================
+
+
+=====   GUNSLINGER   =====
+[Insert cool tasty text about John Woo]
+Class Primary Attribute: Perception
```
Turtlelord26: I strongly advocate for Dex being primary for gunslingers.

play/PlayerRules/CharacterCreation/02_Classes.txt
```diff
@@ -2,645 +2,696 @@ Core classes:
...
-
-"Alchemist"
+	----------------------
+	Elemental Strike -
+	Choose a type from the following list, and deal that type to your
+	opponent on each successful unarmed strike, with its listed effects:
+		Fire: 10 burn damage + 1x Burning.
+		Ice: 10 cold damage + 1x Frozen.
+		Plasma: 10 Plasma damage + 1x Plasma Burn.
+		Electricity: 10 electrical damage.
```
Turtlelord26, line 654: Plus short out a system?

play/PlayerRules/CharacterCreation/02_Classes.txt
```diff
@@ -2,645 +2,696 @@ Core classes:
...
+		Necrotic: 15 necrotic damage + 1x Necrosis
+		Vampiric: Heal self for 10, target loses 20 Nanites.
+		Capacitive: target loses 20 Nanites and receives 2x Plasma Burn.
+		Other [Discuss any ideas you have with the GM).
+	----------------------
+	Shadow Strike -
+	Instead of rolling to hit based on opponent Guard DC + their Skill, you 
+	only have to pass a DC 50 Knowledge [Nanite] check.
+	----------------------
+	Shadow Walk - Requires Shadow Strike
```
Turtlelord26, line 676: Why not use our normal
|
|
V
prerequisite indicator?

play/PlayerRules/CharacterCreation/03_Feats.txt
```diff
@@ -457,76 +461,169 @@ Proficient: Explosives
...
+			up for up to +20 hitpoints. If >5 heal one wound for up to +30 hitpoints.	
+	Charisma = 12
+			Twice per encounter, once per turn if you aren't knocked out, as an immediate 
+			reaction to being at or below half health, roll 1d10. If  <= 5, heal one wound
+			up for up to +25 hitpoints. If >5 heal one wound for up to +40 hitpoints.
+	Charisma = 14
+			Twice per encounter, once per turn if you aren't knocked out, as an immediate 
+			reaction to being at or below half health, roll 1d10. If  <= 5, heal one wound
+			up for up to +25 hitpoints. If >5 heal one wound for up to +40 hitpoints. In 
+			addition, purge 1 debuff.
```
Turtlelord26, line 625: Mildly concerned by 'purge 1 debuff' with no qualifiers or restrictions. Not that I'm going to try arguing that permanent blindness is a debuff, but there are horror stories in Pathfinder of a maneuver known as Iron Heart Surge.

Turtlelord26 (commented):
I did not look at anything in weaponry nor did I look at feats for balance. Lots and lots of notes in classes though.

Fix requests in individual notations. *Probably* nothing that explicitly blocks the merge, but addressing them now would save me creating a few github issues.

Overall looks like excellent stuff. Confident feeling going into initial playtests. 👍

play/PlayerRules/CharacterCreation/02_Classes.txt
1sourcecontrol: How come?

play/PlayerRules/CharacterCreation/03_Feats.txt
1sourcecontrol: Lolol Iron heart surge is pretty funny. Good catch, thanks!

play/PlayerRules/CharacterCreation/02_Classes.txt
Turtlelord26: Because manual dexterity is more vital to shooting than the ability to see fine details. I can see a target downrange in 1080p, but I can't hit it if I can't hold my weapon steady. Particularly regarding the gunslinger, you don't need superior perception to know to shoot people-shaped targets center of mass (or head), but you do need to be able to point that pistol in the right direction *under massive time pressure* [points at gun-kata]. So, because I see the nature of gunslinger ability usage as revolving around fast-paced gunplay, Dex seems more appropriate.
I do not apply this argument to the marksman because even though they still need to shoot (and that shooting should rely on dex), their ability usage seems flavored more towards deliberate, precise action in the right place at the right moment, with emphasis on picking the right target in the first place, and from much, much farther away.

One of the instructors at my college archery club once related a story about an Olympic archery competitor - who had such poor eyesight that he was legally blind. Because good form (physical stances and motions) is so much more important than anything else when shooting.

play/PlayerRules/CharacterCreation/02_Classes.txt
bleehu: Not here. That is a separate issue that we should fix atomically with another PR. This is about Combat 2.0.

play/PlayerRules/CharacterCreation/02_Classes.txt
```diff
@@ -2,645 +2,696 @@ Core classes:
...
+	Your expertise with heavy weapons is overkill for the task at hand. 
+	Gain -3 Miss Chance and a bonus +35 damage for an attack with a heavy weapon.
+	(This miss chance bonus can exceed the minimum -2 ranged miss chance)
+======================================
+
+
+=====   DOCTOR   =====
+You know how or why something works [most likely humans / humanoid bodies), 
+and put that knowledge to good use by supporting your team.
+Class Primary Attribute: Dexterity
```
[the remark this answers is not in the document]
bleehu: That's a great suggestion to use Dex Mod in Medic Spells! Can you create an issue and assign it to me?

play/PlayerRules/CharacterCreation/02_Classes.txt
```diff
@@ -2,645 +2,696 @@ Core classes:
...
+	----------------------
+	Heightened Alertness -
+	Being alert comes with your level of training. Gain 10 Reflex. You cannot be caught 
+	flat-footed (you cannot be penalized for being surprised), but opponents can still 
+	get surprise rounds against you.
+	----------------------
+	Protective Awareness - Requires a minimum Reflex of 24
+	Assiduously observant, you stay aware of any attacks made against you or allies. 
+	Whenever an enemy attempts to make an opportunity attack on you or any allies within 
+	eye-sight, you first gain an attack on them.
```
bleehu, line 469: As a free action?

play/PlayerRules/CharacterCreation/02_Classes.txt
```diff
@@ -2,645 +2,696 @@ Core classes:
...
-Weapon Proficiencies: Pistol, Hand to Hand Expert
-Armour Proficiencies: Concealed, Recon, Light, Cloaking.
-Starting Wealth: $1700+100*Luck - 1300
+	----------------------
+	Heightened Alertness -
+	Being alert comes with your level of training. Gain 10 Reflex. You cannot be caught 
+	flat-footed (you cannot be penalized for being surprised), but opponents can still 
+	get surprise rounds against you.
+	----------------------
+	Protective Awareness - Requires a minimum Reflex of 24
```
bleehu, line 466: 24 is a really really high reflex. It's a really high bonus. Just curious about the thinking here.

play/PlayerRules/Basic Rules.txt
1sourcecontrol: Well it does relate to carrying. Only heavier weapons and armor will restrict your movement (due to weight)

1sourcecontrol: Oh, and the reason I said "not carrying in the traditional sense" is to denote that this system is different than, "I have 5 carrying ability so I can carry 50 kg"

Turtlelord26: I see your point but I would still prefer to find something we can call it that doesn't require a clarifying statement in the first place.

play/PlayerRules/Basic Rules.txt
1sourcecontrol: If we do want the engineer to be better at repair than everyone else, I think we should just give him a class bonus that doesn't count towards the 50 point max. I think a growing cap based on level could be reasonable. Do you want to open an issue?

bleehu (approved):

play/PlayerRules/CharacterCreation/02_Classes.txt
1sourcecontrol: Gives more value to Heightened Alertness, which adds 10 reflex.
Also, Protective Awareness is extremely powerful, so having a high minimum will either force the player to min-max very hard or wait until level 10 so they can first acquire Heightened Alertness.

1sourcecontrol: Oh, and the player will likely be vying for high reflex already since that gives them a passive miss-chance defense. So I think it's prudent to set the minimum to be quite high so the ability isn't a "gimme" but requires more thought

play/PlayerRules/CharacterCreation/02_Classes.txt
1sourcecontrol: Yeah, I need to mention that fact that their weapon range is a limitation.
And on further review, it'd probably be prudent to put in some sort of cap -- maybe from player level, a hard cap, or nanite cost

1sourcecontrol: Also, my previous logic also includes the fact that the gunslinger is stuck with pistols which don't scale well into the late game against armored targets (unless using lasers, which have their own limitations such as a lower average base damage).

play/PlayerRules/CharacterCreation/02_Classes.txt
```diff
@@ -2,645 +2,696 @@ Core classes:
...
+	The Trail becomes bright against a monochrome background of edged shades of 
+	grey. You can see marks and paths that others simply cannot. +50% to a tracking 
+	check, or maximum 2 turns to pass. Passively, gain the ability to sense where your
+	Marked Quarry is, within 20 meters of you.
+	----------------------
+	Highlight Quarry - 
+	If your Quarry is marked, for 1 action, you can activate a Marked Quarry's nanites to make them 
+	more visible. All damage dealt to this target is increased by 10. Two uses per encounter, one use
+	per Quarry. Maximum of Lvl number uses per day. Lasts for the duration of the encounter.
+	After level 10, all damage dealt to this target is increased by 20 (instead of 10).
```
1sourcecontrol, line 313: Note to self, "at level 10" instead of "after level 10"

play/PlayerRules/Basic Rules.txt
1sourcecontrol: Yes, after a discussion with Taylor we both thought it was more prudent to use our existing resource pool (nanites) instead of adding a secondary one and splitting abilities between the two pools. Reasoning: less complex. Although in retrospect, having Nanites separated out as something that only activates abilities that are super-human would be a way to devalue nanites. Which might mean that using a class primary attribute system for stamina or nanites might be more feasible.

Or.. that gives me an idea. What if we attached a stamina resource/attribute to FORT points, and made all the athletic active feats use stamina instead of Nanites? That would significantly improve FORT's value.

Turtlelord26: 1. The extreme example is medic, who we probably don't want to have to manage health, stamina, nanites, adrenaline, *and* contagion resources.
2. We would still run into all of the design problems we've talked about with tying health and a resource to the same stat.
3. Also we would be indirectly hurting characters who never take athletic feats as part of their aesthetic.
So, I am currently in favor of having one general ability-focused resource pool.

play/PlayerRules/Basic Rules.txt
bleehu: Makes sense. but this is not a blocking issue. This is a new feature that doesn't have to happen in C2.0. Shipit and we'll do this work next.

play/PlayerRules/CharacterCreation/02_Classes.txt
1sourcecontrol: @bleehu

play/PlayerRules/CharacterCreation/02_Classes.txt
1sourcecontrol: @bleehu

play/PlayerRules/CharacterCreation/02_Classes.txt
bleehu: Thanks for letting me know the thinking. All good.

play/PlayerRules/CharacterCreation/02_Classes.txt
bleehu: Combat 2.0 is already too bloated as it is.

play/PlayerRules/CharacterCreation/02_Classes.txt
bleehu: It helps make it obvious what the difference is from the baseline, but I've always thought it was weird that we didn't simplify it also.

bleehu: However, once again, this is a separate non-combat 2.0 specific issue.

play/PlayerRules/CharacterCreation/02_Classes.txt
bleehu: I think we can improve this, but this is not a blocking change. We can ship 2.0 and then nitpick this one feat. That's why my review said approved. 
I would like to see this interrupt cost an action on the following turn or something. I kinda think that if the gunslinger can sometimes aim and fire without it costing an action, then players will ask or subconsciously think "then why does it cost me an action to aim and an action to fire normally? Is it just because it's a game?" I want to be careful about trying to nuke everywhere where that sentiment crops up, because we'll end up going around in circles again about the balance between fun-realism, but I think there's enough cognitive dissonance here to ask for some cost besides just a really high reflex.

play/PlayerRules/Basic Rules.txt
1sourcecontrol: Fair enough. We could try and remove the clarifying statement and state simply how the system doesn't work like a strength = kg system

play/PlayerRules/Basic Rules.txt
1sourcecontrol: kk

play/PlayerRules/Basic Rules.txt
1sourcecontrol: Good points, fair enough.

play/PlayerRules/Basic Rules.txt
1sourcecontrol: we'll yolo it and see if it breaks.

play/PlayerRules/Basic Rules.txt
1sourcecontrol: kk

play/PlayerRules/CharacterCreation/02_Classes.txt
1sourcecontrol: thx man :D

play/PlayerRules/CharacterCreation/02_Classes.txt
1sourcecontrol: It doesn't do a tremendous amount when auto weapons already fire ~6 shots. Although, firing an extra shot in semi auto is super powerful.....

play/PlayerRules/CharacterCreation/02_Classes.txt
1sourcecontrol: @bleehu , thoughts?

bleehu: I don't want to short circuit the dynamic that the primary source of adrenaline should be from helping teammates. Keep in mind, Harmacist spike only costs 30 nanites. So if someone misses 3 shots with an SMG in one action, they've charged that ability up entirely.

Issue #133, opened 2017-07-20
Death saving throws are a house rule


bleehu:
Without some bleedout system, there's no actual danger from going down. We should make the death saving throws part of the core rules. If we don't do that or implement a similar system, then we'll suffer the cockroach effect.

trowl223:
I'm going to quote out of D&D 5e because I like the way they handle health points being reduced to 0.

As taken from Roll20.net:

> Death Saving Throws
> Whenever you start your turn with 0 hit points, you must make a Special saving throw, called a death saving throw, to determine whether you creep closer to death or hang onto life. Unlike other saving throws, this one isn’t tied to any ability score. You are in the hands of fate now, aided only by Spells and features that improve your chances of succeeding on a saving throw.
> Roll a d20: If the roll is 10 or higher, you succeed. Otherwise, you fail. A success or failure has no effect by itself. On your third success, you become stable (see below). On your third failure, you die. The successes and failures don’t need to be consecutive; keep track of both until you collect three of a kind. The number of both is reset to zero when you regain any hit points or become stable.

https://roll20.net/compendium/dnd5e/Combat#content, D&D Fifth Edition Compendium, Wizards of the Coast, LLC.

The system I like best for Compound_X is the following:
When you have 0 Hit Points, your Actions per Turn are reduced to 0 and you must make a Death Saving Throw once per turn. The maximum number of saving throws you can make per turn is half your usual Action allowance, rounding down (4 for most players, I don't think we've added the ability to increase the number of Actions per Turn yet, have we? I'd like that to be a thing with certain aspects, like having auto-firing mounts on Combat-Frames, for example. I digress . . .) The DC for this saving throw is equal to 70 - any bonuses to their Knock-out Threshold, including any buffs to this stat by any abilities or effects (Gives us room for some spells and abilities here). Any Cyborgs with Internal Reinforcing, for example, would have a DC of 65 since their K-oT is increased by 5.
On success, the player becomes conscious. The player has a number of chances to succeed equal to the average of their 3 physical stats (Fortitude, Strength and Dexterity), rounding down. Once all chances have been consumed, the player is evicted, so to speak.
The chances are replenished once the player is able to take a number of long rests equal to 10 - the average of their mental stats (Perception, Charisma and Inteligence), rounding up. This period resets whenever the player has to make a Death Saving Throw in combat.

The differences between my proposed system and the one used in D&D 5e, a relatively well balanced game, as I perceive them, are the following:

[...]

PR #171, opened 2017-10-30
MH-Basic-Rules: Death and Dying

bleehu:
This time merged into Dev, not play. Intended to fix Issue #133


1sourcecontrol (approved):
I'd say go for it. The only thing I'm thinking about is the new "capital letter for game terms" rule we discussed. You could do that now or or not worry about it.

PR #172, opened 2017-10-30
Mh polish play

bleehu:
One small step toward being able to let new players just read the play branch


1sourcecontrol (approved):
Lit like a light

PR #175, opened 2017-10-30
removed Taylor's old character sheets

bleehu:
People's actual character sheets should be stored in Google Drive or something, not the core rules repository.


1sourcecontrol (approved):
hit it!

PR #179, opened 2017-10-31
MH Last Stand

bleehu:
I screwed up and merged #157  early. The concern was that the feat was underpowerd, so I removed the nanite cost and miss chance.


Wildlovelies:
my concern is, if they spend 40 nanites to draw their weapon, then pass back out because they are down, it can be a useless feat, unless they can use it in succession. but it would take 3 uses to draw a weapon aim and fire, thus 120 nanite cost, yes they can fire without aiming, but with 2.0 in mind where accuracy is incredibly difficult to get 0% malice, you are looking at a single shot with a minimum of a 7, most likely 8 to hit. thus a character has to roll a 9 or 10 to even hit, can "free actions" be used? (quick draw, twitch, slings, etc)

bleehu:
I have removed the nanite cost entirely from the Last Stand feat. Please note the obvious problem if Last Stand does not coexist with the change in the rules allowing an enemy to Coup De Grace.

Wildlovelies:
i propose a re-write. 

instead of last stand, I propose: "Not Dead Yet!"

Not Dead Yet!: 
Reqs survival 20, fort 5

After each successful passing of the fort save for the "downed" condition, the player may pick one of the following: speak normally, draw a one handed weapon (pistol, smg, one-handed melee weapon, or sawed off shotgun), aim, use equipped weapon, or crawl 1m. Free actions may be used in conjunction with this feat.

CharacterCreation/03_Feats.txt
```diff
@@ -190,6 +190,11 @@ lower chance to hit than the previous hit. (A 4th level character making their
 applies to a monk's Shadow Strike as well, increasing the DC by 20 per strike 
 instead of increasing the strike's miss chance.
 
+Last Stand:
+Prerequisite: Level 4
+Don't die alone. Always bring a friend. You may take an action to do one of 
+the following: Draw a one-handed weapon such as a pistol or SMG, Aim, or fire.
```
1sourcecontrol: @bleehu I hope you won't lambaste me for saying this since we've had recent drama about writing styles (I have good intentions):
I think it would be more clear to the reader if the latter half of the last sentence gets rewritten from: 

>Draw a one-handed weapon such as a pistol or SMG, Aim, or fire. 

to

>Draw, Aim, or Fire a one-handed weapon such as a pistol or SMG.


Also, do you intend for the feat to allow you to take one action per turn or just one action per Down, **period**?
Also, you removed the part that explains what triggers this ability.

1sourcecontrol (changes requested):

CharacterCreation/03_Feats.txt
Wildlovelies: "After each successful passing of the fort save for the "downed" condition," is what i wrote for the trigger, and going off of basic rules: " You roll one saving throw per turn 
that you are downed and these DSTs are not modified by any attribute" 

Meaning if you succeed your death saving throw and have this feat, you can do something small as defined in the feat instead of just losing your turn.

Wildlovelies: change fort to dst, maybe add the only dst modifier to it? maybe add 10% making a roll of 40%?

Wildlovelies: would have to update basic rules.

1sourcecontrol:
@bleehu lol you never explained to me what you meant with "to-miss penalty of 2"! When I read it a few times, I thought it meant "this to-miss chance penalty of 2 replaces your weapons normal miss chance". I think I finally understand that you meant a miss chance penalty in addition to normal miss chance. 

P.S. I think it was a good idea to remove the nanite costs and the miss chance. Pending your response to Lovelies and my concerns (noted in the review) I think it's great.

@Wildlovelies don't forget that survival is about surviving in the wild and making camp fires and stuff, i'm not sure if it's relevant to surviving bullet wounds. I think your rework thoughts have a lot of merit! (except for the 1m crawl: there already is a feat in the works that does movement while Downed). Though, I'll let @bleehu decide the direction he wants to take the feat.

Wildlovelies:
@1sourcecontrol In the past, survival has been used with fort for death saving throws (house ruled), initially i was commenting using play branch (which says fort save for dst) + previous experience. to which is not in the rules. I was also not aware of the movement while downed feat, i must have missed that. Honestly i think it would be best to combine them. use this feat for action while downed.  Having both could cause some balance issues and zombie style characters without the coup de grace rules initiated in the game. Adding movement to this feat and removing the others means if you succeed your dst then instead of just ending your turn you can do something, not much but better than nothing. that is the spirit of my idea

bleehu:
Coup De Grace, by another name is implemented in dev: https://github.com/bleehu/Compound_X/blob/dev/Basic%20Rules.txt#L474
As such, I'm really not too concerned with a Zombie Build, if it means that one good enemy SMG strafe sends a player to oblivion. I like the desperate crawl being separate from the Last Stand; I feel like it separates the two most tactically advantageous maneuvers which lend themselves to a hail Mary play; moving and shooting. Having those balanced but separate I feel gives the player agency/meaningful choices.

1sourcecontrol:
@bleehu 
When you merged it early, I see that:

> "When downed, any additional attack that hits regardless of damage is the same as
having failed one Death Saving Throw."

Was merged as well. As stated in #157  I'm still of the opinion that this will be catastrophic and should be altered to something like I recommended in #157.

CharacterCreation/03_Feats.txt
1sourcecontrol: @bleehu please respond to my questions, above.

bleehu:
Wouldn't -50 health be worse since most rifles deal more than 50 damage!?!

CharacterCreation/03_Feats.txt
```diff
@@ -531,12 +537,12 @@ their melee weapon(s). The attacks must take place within a 5m space, rolling fo
...
-For 1 Action, reduce weapon range to 15m, and negate actions needed for Aiming, Pumping,
-and Cocking of a gun. Treat your gun as if it was a CQB weapon. After a successful 
-hit, you may roll for an additional hit on any target within the 15m range, for up to 
-3 rounds fired. If you miss a shot or your gun runs out of rounds, your Gun Kata ends.
+Gun Kata [30 Nanites / Turn]: 
+Prerequisites: Level 10 or higher
+For up to 3rds, reduce weapon range to 15m, reduce Miss Chance by 50%. Treat 
+any weapon as if it were a CQB weapon. After a successful strike, you may roll 
+for an additional strike during the same action on any target within the 15m 
+range.
```
1sourcecontrol: Please fix this old gun kata.

CharacterCreation/03_Feats.txt
```diff
@@ -520,7 +526,7 @@ to the beginning of your next turn. Cannot be used in concert with Run 'N' Gun.
 
 Cover Slide [15 Nanites]:
 Prerequisites: DEX 4
-Move up to your speed and end next to a barrier behind which you can take 
+Move up to your speed and end next to a barrier behind which you can take
```
1sourcecontrol, line 530: Please put the space back, I was using it.

CharacterCreation/03_Feats.txt
```diff
@@ -200,6 +200,12 @@ head out and looking at your enemy. And somehow you still hit them (some of the
 Roll a d10 before any Blind Firing or Hip Firing action. If the roll is from 6 to 10, 
 Gain -2 miss chance for that firing action; If the roll is from 1 to 5, nothing happens.
 
+Last Stand:
+Prerequisite: Level 4
+Don't die alone. Always bring a friend. You may take an action to do one of 
+the following: Draw a one-handed weapon such as a pistol or SMG, Aim, or fire.
+You can no longer take any of these actions if you are below -50 health.
```
1sourcecontrol: Per our conversation.
Please clarify the writing:
Do you intend for the feat to allow you to take one action per turn or just one action per Down, period?
Also, you removed the part that explains what triggers this ability.

Basic Rules.txt
```diff
@@ -489,8 +489,8 @@ they cannot move, cannot stand under their own power, and fall prone if not
 supported by other means. They cannot speak, but can mumble incomprehensibly.
 Every turn that a character takes while they are downed, they must make a Death
 Saving Throw of DC 50. Success and failure of these throws is detailed above.
-When downed, any additional attack that hits regardless of damage is the same as
-having failed one Death Saving Throw.
+Even if you had 5 health and took 80 damage, when you go down you go to 0 health.
+Any additional attacks you sustain will push you into negative health.
```
1sourcecontrol, line 493: Should you mention that there are no penalties (besides not being able to use certain feats) for getting lots of negative health? up to you

1sourcecontrol (changes requested):

CharacterCreation/03_Feats.txt
1sourcecontrol: kappa

bleehu: kappa? as in just K? I can't tell if you are serious or not.

1sourcecontrol: I'm totally joking haha. "kappa" is an expression coined by Twitch chat, usually used for comedic effect

bleehu: lol! Thanks for the explination. I waz teh confuz

CharacterCreation/03_Feats.txt
```diff
@@ -200,6 +200,12 @@ head out and looking at your enemy. And somehow you still hit them (some of the
 Roll a d10 before any Blind Firing or Hip Firing action. If the roll is from 6 to 10, 
 Gain -2 miss chance for that firing action; If the roll is from 1 to 5, nothing happens.
 
+Last Stand:
+Prerequisite: Level 4
+Don't die alone. Always bring a friend. You may take an action to do one of
```
1sourcecontrol: Please add some mention of how Last Stand is triggered (just to be crystal clear). Also, please add how many actions per turn they get since being "Downed" normally means you don't get any actions except saving throws (although that isn't stated specifically in the "Down" section of the basic rules). I.e., "If you are downed you may take 4-actions per turn to do the following (in addition to rolling your saving throw):"

1sourcecontrol (changes requested):
One/two last changes, then I think you're good to go!

bleehu:
Let's do one last review just to make sure. Gadzooks, 22 days for one feat. yuck. 

I figured three actions seemed reasonable to reflect that the player is in a weakened state (since they're working on bleeding out) and since they don't have the option of hunkering; being prone gives some bonuses to long range defense anyway. 

Super chill?

1sourcecontrol (approved):
chill super!
(btw, where did that "super chill" saying come from?)

bleehu:
I haven't the slightest idea where I picked that up. Shippin' it.

PR #180, opened 2017-11-01
Roll with Advantage Terminology in Lucky Feat

Turtlelord26:
This is what Lucky looks like with Advantage terminology.

It only really works with the 8 and 9 cases. Is that enough to justify it here?


bleehu (changes requested):
In order to call this Advantage, we should define Advantage in Basic Rules under Primary Game Mechanic. 
https://github.com/bleehu/Compound_X/blob/play/Basic%20Rules.txt#L50

bleehu (approved):
shipit, baby!
https://youtu.be/h-drS62OnrQ?t=13s

Issue #183, opened 2017-11-07
Wealthy sponsor added back into play branch


Wildlovelies:
Wealthy Sponsor: 
Prerequisites:
Gain LUK * Lvl * 750 credits.

PR #188, opened 2017-11-07
Apply hotfixes to R_C

bleehu:
So that we stop re-adding content that we needed to yank. Should also help keep #183 fixed.


1sourcecontrol:
shipppppppppppppppppppppppppppppppppppppppppppppppppppit

1sourcecontrol (approved):

PR #197, opened 2017-11-16
Version Beta 0.0.1

bleehu:
With the inception of Combat 2.0, I'd like to change the versioning scheme to have 'Beta' before it. We'll continue to bump the second number with major feature releases, the last number with every hotfix. 
The first version number will go from 0 to 1 when we publish. 

While it would be nice for Reveraine to weigh in, This merge is pending the approval of 1sourcecontrol.


1sourcecontrol (approved):
I didn't look over all the C2.0 stuff again, since nothing changed from the initial merge. I did see the glossary, however. 
Also, you'll be merging without my patch to remove dev notes and gunslinger balance. Just BTW

And btw, Idk what I'm supposed to do here, like, what is my role for these merge to R_C PRs? Is it more of a formality? Most of this information I've reviewed or proposed myself already.

And last 'btw', if you want to wait for Sarah to review, I'm 100% fine with that.

Reveraine (approved):
Yay renaming things! =)

bleehu:
@1sourcecontrol, the redundancy there is coming from you wearing both the Director hat and the Lead Developer Hat. This is one of the several spots that you as the director get final veto power over what goes into your game. It's more than a formality, but shouldn't be a big fanfare. It's kinda nice having both directors and the producer all on the same page shipping a new version of the game to R_C.

Issue #174, opened 2017-10-30
Class Feat Acquisition Rules Ambiguity


Turtlelord26:
In 02_Classes.txt

"you do not select a class feat until level 5"
"At the first level taken in a class // (other than level 1), you get the single star version of the feat"

Are quoted out of the same block of text. My understanding is that there is general consensus on the first quote. We should in that case delete the second.

PR #193, opened 2017-11-15
C2.0 fixes to dev

1sourcecontrol:
Fixed an ambiguity in the classes doc.
Merged class starting money numbers.
Nerfed "Needs More Dakka" nanite cost to 15.
Altered Gunslinger's "Protective Awareness" Feat
Removed Mage and Sorcerer.
Buffed Monk so that it's a little more playable in C2.0 -- I think there is potential to have some kind of Zealot class, that at least has more purchasing power than the existing monk and actually fits our aesthetic. I didn't want to yank Monk because it's actually playable, incase anyone wants to, until we redo the classes.
Various text fixes.
Removed my dev note from "Rainman" and buffed rainman -- tell me what you think!
Clarified text in "Positional Mastermind" and gave it a slight nerf (to prevent abuse by gunslinger).
Modified "Will to Survive" to prevent purging death and whatnot.

PR #199, opened 2017-11-16
Fix Issue #174.

bleehu:
clarified ambiguity in class feat acquisition elucidated in Issue #174. Please note added reference to Skills in Basic Rules doc instead of having a rule oddly stuck in the Classes doc.


Turtlelord26 (changes requested):
Skills doc reference noted and appreciated.

New text doesn't scan very well. We have line 6 referencing class feat initial acquisition, then talk about skills before we get back to class feats. 
Would like to move skills wholly before or after class feat text. 
Also the reference to "level 5 ..." in line 10 should be a reference to "level 10 and every 5 levels thereafter" or just "levels 10, 15, and 20".

1sourcecontrol:
So, I don't understand. Are we doing this change twice #193 because I forgot to link the issue, or do you want to redo what I was PR'ing?

bleehu:
@1sourcecontrol If I'm seeing this right, the two changes are next to each other but not the same? This change is for class feats, the one I requested in 193 is for the sentence about skills.

1sourcecontrol:
image

We state a lot of the same things, just in two different locations.

1sourcecontrol:
I'll save us more trouble and just merge my stuff, then you can figure out the conflict however you like

bleehu:
Especially interested in 1Sourcecontrol's feedback on language at the top. I don't want to squash your work but I think I did this time. Let me know if you like your way better.

1sourcecontrol (approved):
can you put the line 11 "RP / Using Skills" section in quotes or something? It may be confusing to the player if we don't indicate that the text is in one block (FYI it was a little confusing to me).

Otherwise, looks great! I am more concerned with the "at level 5 gain class feat" stuff, and that looks good now. So we kool.

Feel free to ship if you agree with me about the above and add the necessary change! :)

1sourcecontrol:
>can you put the line 11 "RP / Using Skills" section in quotes or something? It may be confusing to the player if we don't indicate that the text is in one block (FYI it was a little confusing to me).

Michael, you merged without making the change or saying anything against it!

1sourcecontrol:
@bleehu

PR #210, opened 2017-11-22
added Multipurpose Drone

bleehu:
Ever notice how we don't actually have an Anatomy section?


Reveraine:
I think Anatomy in a base sense is very important. Because we as humans assume the head is the weak point, but for droids, that is does not necessarily have to be true.

1sourcecontrol (commented):
Wouldn't being Faceless make it harder to deceive/persuade/etc. due to not having convincing facial expressions (giving rise to doubts of the sincerity of the robot?). This question assumes less people have repair skill than a simple higher than 5 charisma. Thinking out loud

I think the drones may be stronger than humans because of the traits "Hovering" and "Faceless", but I think it's close enough to warrant merging and testing. Especially since humans have better stat malleability.

Can we rename Drone to Robot? Drone implies to me that the robot is controlled remotely. (Unless that is the intent. But from your "Society" it looks like the drones are self-controlled.)

Turtlelord26:
Concur, drone definitely implies remote operation. Alternatives include, robot, construct, golem, etc.
Incidentally, can they make personality backups ~~like archons~~?

bleehu:
I kinda hate the name of it overall. I think more improvement could be done on more than just the name. Note that the other robotic race is also called a Security Drone. Should we change it in both places?

The balance to "Human" and "Faceless" was supposed to be the lower number of stat points. 

I'd like to get some suggestions before we merge and test. And I'd also like to confirm that adding another robotic race is a good idea in general (I didn't coordinate with anyone before submitting this). I wanted to use a specific example of adding a new race before I just lobbed the idea. 

The thought was definitely that they weren't remotely controlled.

1sourcecontrol:
I think we need to change it in both places.

I understand. 

I think it's a great idea. I'm not sure I have any good ideas... but you could put some limitation on hovering (like nanites or uses per encounter) .. but then the robot needs legs or wheels.

bleehu:
Should we hold on this until after the existing Races rework?

rosebrallier:
We can certainly hold on to this and see how it meshes with the work we do on the security bot.

CharacterCreation/01_Races.txt
```diff
@@ -109,6 +109,46 @@ creation. This is considered a Class skill.
 Immune to: poison, cannot be intoxicated.
 Weakness (Electrical): Electrical damage deals an additional 50% damage.
 
+//============================================================
```
1sourcecontrol: Remove "//"?

1sourcecontrol:
I'd be fine with merging it. That way we can play it if the race rework takes a long time. :D

bleehu:
Alrighty! It's fun work and I'll do my best, but it's not super high on the priority list.

bleehu:
Gonna re-do this. It's old enough that almost all of it has to be re-done.

PR #212, opened 2017-11-23
modify push/shove

bleehu:
because it's fun to nitpick!

### DIFF
--- a/Basic Rules.txt
+++ b/Basic Rules.txt
@@ -643,7 +643,7 @@ Instead of dealing full damage, a character can try to kick/shove a human sized
 opponent to force movement by rolling for a normal unarmed/armed melee attack 
 with an additional -2 miss penalty against the opponent's Guard DC. If the attack 
 succeeds, it yields 1/2 the damage it normally would and pushes the target 
-1m for every 5 strength you have.
+1m for up to every 5 strength you have.
 
 === Actions ===
 During combat, your character has 4 Actions. Actions are used for fighting, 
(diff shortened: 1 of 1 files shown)


1sourcecontrol (commented):
Can you explain your reasoning?
Isn't it intuitive enough for "1m for every 5 strength"?

FYI the intended rule is that 0-4 STR yields 0m of forced movement. 5-9 is 1m, 10-14 is 2m, and so on.

bleehu:
The way it's written, if you have an STR higher than 9, you cannot push 1m, you _must_  push 2m, which might not be as advantageous

Basic Rules.txt
```diff
@@ -643,7 +643,7 @@ Instead of dealing full damage, a character can try to kick/shove a human sized
 opponent to force movement by rolling for a normal unarmed/armed melee attack 
 with an additional -2 miss penalty against the opponent's Guard DC. If the attack 
 succeeds, it yields 1/2 the damage it normally would and pushes the target 
-1m for every 5 strength you have.
+1m for up to every 5 strength you have.
```
1sourcecontrol: Can you add a pair of commas or parenthesis around "up to"? The prepositional phrase, or whatever that is, isn't 'essential to the meaning of the sentence,' and would probably be easier to read with additional help.

1sourcecontrol (changes requested):
somethin' smol

1sourcecontrol:
Good catch, btw

Basic Rules.txt
Turtlelord26: The fix you want is to move the 'up to' before the '1m', so you have 'up to 1m per 5 strength'. Grammatically it cannot be turned into an appositive because it isn't a clause.

bleehu: Can do! :)

Turtlelord26 (approved):
👍

1sourcecontrol (approved):
Grape!

PR #219, opened 2017-11-27
Basic Rules conformity to the style guide and other

1sourcecontrol:
- Edited out colons in some chapter titles... I think it looks cleaner.
- Made styles more consistent by increasing # of "=" signs in section declarations.
- Also, scope creep, I noticed that a definition for Nanites never got established so I added some text. (I know there is some flux about this, but let's put something there until further notice).
- Fixed some text in the cloaking section
- Reworded some text in combat stat examples

### DIFF
--- a/Basic Rules.txt
+++ b/Basic Rules.txt
@@ -1064,7 +1064,7 @@ sooner (see table below for "Cloak Turns") . If your cloak time runs out, you st
 cloaking and have to wait a number of actions equal to your "Charge Delay" (listed under 
 the relevant cloaking module's description) before being able to cloak again.
 
-===== Cloaking Detection ===
+===== Cloaking Detection =====
 If someone/something is looking for you, they must pass a (normally, perception) 
 check at 'Detect DC' difficulty, modified by your stealth skill. While moving, 
 the detect DC is reduced (listed as 'moving detect DC' under the relevant cloaking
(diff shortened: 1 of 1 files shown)


Basic Rules.txt
```diff
@@ -117,9 +115,14 @@ Class 		- Think of this like your job. Are you a sniper, waiting for the
...
-	money you start with to buy equipment and how many nanites you can use
+	money you start with to buy equipment and how many Nanites you can use
 	at once.
-Nanites  	- 
+Nanites  	- The evolution of technology brought about the ability to design
+	and control microscopic machines. Such a powerful advancement assists in
+	all forms of life in the Compound X universe. Nanites are used to activate
+	abilities that are not normally possible on Earth. Abilities such as sprinting 
+	extremely fast, aiming incredibly well, infusing bullets with electrical
+	energy, and more.
```
bleehu: While I'm normally quite loathe to accept scope creep, I think you handled it pretty well here. I do think "not normally possible _on earth_ " is a little non-diajetic and I'd like to see if we can tweak it, but that's a nitpick. Great examples!

bleehu (approved):
shipit!

Basic Rules.txt
1sourcecontrol: I'll try and change it to something more in-game. I noticed that other areas seemed to break some of the '4th wall' if you will, so I was trying to build some excitement by comparing what's possible on earth.

bleehu:
Looks good! Shipit! Great work, 1Source!

Issue #224, opened 2017-11-28
Add distance limit to prone


1sourcecontrol:
As mentioned in #166 , add a distance limit for defensive bonuses while prone. I think the bonus should still be given to crouch at any distance, since you are a smaller target.

PR #225, opened 2017-11-28
Modify prone rules

1sourcecontrol:
per issue #224

### DIFF
--- a/Basic Rules.txt
+++ b/Basic Rules.txt
@@ -842,7 +842,7 @@ either by partially obscuring your silhouette, or by providing slight protection
 from projectiles. Full cover will do both to a better degree. "Partial Cover" typically
 covers 30-50% of your body, whereas "Full Cover" will cover closer to 60-80%. Crouching
 or going prone will also take better advantage of your surroundings and make you a 
-smaller target. Debuffs to opponent's Miss Chances are as follows:
+smaller target. Debuffs to opponents' Miss Chances are as follows:
 
 No Cover: +0
 Partial Cover: +2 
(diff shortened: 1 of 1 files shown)


Basic Rules.txt
```diff
@@ -842,14 +842,16 @@ either by partially obscuring your silhouette, or by providing slight protection
 from projectiles. Full cover will do both to a better degree. "Partial Cover" typically
 covers 30-50% of your body, whereas "Full Cover" will cover closer to 60-80%. Crouching
 or going prone will also take better advantage of your surroundings and make you a 
-smaller target. Debuffs to enemies Miss Chance are as follows:
+smaller target. Debuffs to opponent's Miss Chances are as follows:
```
Turtlelord26: Should be "opponents' Miss Chances" as there most often will be more than one. Difference is location of the apostrophe.

Basic Rules.txt
```diff
@@ -842,14 +842,16 @@ either by partially obscuring your silhouette, or by providing slight protection
...
 
 No Cover: +0
 Partial Cover: +2 
 Full Cover: +4
 
 Crouching: +1 (on top of whatever cover you're behind, may completely obscure vision)
-Prone: +2 (on top of whatever -partial- cover you're behind, otherwise, may completely obscure vision)
+Prone: +2 (on top of whatever -partial- cover you're behind, otherwise, may completely obscure vision.
+	    Provides no bonus if an opponent is within 5m of you and you aren't in cover from their
```
Turtlelord26: No bonus at all? Against anyone? Or, preferably,  just no bonus against enemies who are within this range? Request the word "if" be changed to "against".

Turtlelord26 (changes requested):
Particular number used subject to playtesting, but as far as merges to dev goes this seems like a good general idea.
See two inlines addressing clarity issues.

Basic Rules.txt
1sourcecontrol: good catch!

Basic Rules.txt
1sourcecontrol: ah, whoops. Thanks for that!

Turtlelord26 (approved):

PR #227, opened 2017-11-29
Modified Classes to conform to plaintext style guides

bleehu:
Isn't that easier to read? It will also be easier for CXDocs to read...

### DIFF
--- a/CharacterCreation/02_Classes.txt
+++ b/CharacterCreation/02_Classes.txt
@@ -1,25 +1,44 @@
-Core classes:
-/*====================================================
- * Your character starts at level 0. At level 0 you have skills, but no class, 
- * class feats or regular feats.
- * At level 1 you get to select a class, and receive the class passives. 
- * However, you do not select a class feat until level 5. Then, at levels 10, 15, and 20
- * you can gain another class feat.
- *
- * For any listed class, if it says "Gain X Skill at 10" that means you get
- * to add the X skill to your skill list. For more on Skills, see 
- * RP / Using Skills Section of Basic Rules.txt. 
- *==================================================*/
+==============================
+Core Classes
+==============================
+
+	Your character starts at level 0. At level 0 you have skills, but no class, 
+class feats or regular feats. At level 1 you get to select a class, and receive 
+the class passives. However, you do not select a class feat until level 5. Then, 
+at levels 10, 15, and 20 you can gain another class feat.
+
+	For any listed class, if it says "Gain X Skill at 10" that means you get to 
+add the X skill to your skill list. For more on Skills, see "RP / Using Skills" 
+Section of Basic Rules.txt. 
+
 
 =====   SOLDIER   =====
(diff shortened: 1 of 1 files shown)


CharacterCreation/02_Classes.txt
```diff
@@ -1,25 +1,44 @@
...
+* Submachine Guns 
+* Shotguns 
+* Carbines
+* Assault Rifles
+* Battle Rifles.
+
+Armor Proficiencies: 
+* Light
+* Medium
+* Medium-Heavy
```
1sourcecontrol, line 39: Idk. 16 lines to display 3 lines of information. I really don't like all the space; I think it will make it much harder to scroll through the doc to compare classes.

bleehu: 1) This is a public repo, get my name out of there. Get it out now. 
2) If you want to address me, @ me. 
3) I hear you, long files are bad. But by breaking up the information, it actually becomes easier to skim to the part of the info that you're looking for thereby making it easier to find and work on what you're trying to do. For players, they'll be using the CXDocs page which will render each of these into Divs which can then be aligned to be side-by-side. Better yet, if the computer can read and differentiate each class, we may be able to actually add a side-by-side tool. But in order to add that quality of life, we need CXDocs to be able to read what we're writing rather than needing a human-in-the-loop to interpret each change.

Turtlelord26: I concur that spacing the information vertically will make comparison faster despite marginally increased scrolling time.

bleehu:
You would not *BELIEVE* how much trouble line 216 with the non-ascii backtick character is giving me. Came from copy-pasting from Microsoft word, which uses Windows-1252 standard language encoding. God knows it's not the last time it's gonna happen. so I've gotta figure out a way to future proof it.

CharacterCreation/02_Classes.txt
```diff
@@ -512,15 +710,28 @@ Class Feats:
...
+
 Class Primary Attribute: Strength
-Passives: Expert Unarmed
-Weapon Proficiencies: Pistols.
-Armor Proficiencies: Concealed, Recon, Light, Medium, Medium-Heavy.
 Starting Wealth: $2170 + 100*Luck
-Class Feats:
+
+Passives:
+* Gain Expert in Unarmed
```
1sourcecontrol: Can you change this to something like , "Gain usage of melee weapon "Hand-to-Hand [Expert]", disregarding requirement"?

I don't think there is an "Expert in Unarmed" and I believe "Unarmed Expert" was changed.

(If this isn't too much scope creep)

1sourcecontrol (approved):
One potential change, go ahead and merge when you're done

1sourcecontrol:
Looks good though! Nice work.

CharacterCreation/02_Classes.txt
bleehu: wilco

PR #229, opened 2017-11-29
Initial styleguide pass through the used files

1sourcecontrol:
I didn't include Paladin's tomb of spells because I'm not sure how long it will be around.


bleehu (approved):
holy cow! It's a monster! Shipit!
Some of the paragraphs will need to be 4-spaced or tabbed, but I'll handle that in a subsequent pull.

1sourcecontrol:
sounds good.

PR #244, opened 2017-12-07
styleguide conformity; armorlist

1sourcecontrol:


CharacterCreation/05_ArmourList.txt
```diff
@@ -174,329 +174,324 @@ APs (Armor Points):		130
...
 Suit integrity: 10 shots
 
-	=====   LEVEL 10 REQUIREMENT   =====
+=====   LEVEL 10 REQUIREMENT   =====
 
-Kevlar Fiber Environment Suit  $2000
+Kevlar Fiber Environment Suit
+Cost:	$2000
+Suit integrity: 20 shots
 Holds 1 trauma (armor) plate in the front and back. 30% coverage, must purchase
```
bleehu: Paragraph needs an indent

CharacterCreation/05_ArmourList.txt
```diff
@@ -524,7 +519,7 @@ Secondary Module Benefit: Either provides +60 shield strength
 OR +60 recharge rate, +4 fast recharge cost per secondary.
 Decided on purchase.
```
bleehu, line 484: These newlines in the end of the stat block are not going to roll into the line above it. I'm not sure what the best solution for this is.

CharacterCreation/05_ArmourList.txt
```diff
@@ -597,7 +592,7 @@ Sec Module Hard Pts:	1
 Secondary Module Benefit: Provides +1 Cloak Turn
 Bonus, -1 Nanite Upkeep, per secondary module
 
-	=====   LEVEL 5 REQUIREMENT   =====
+=====   LEVEL 5 REQUIREMENT   =====
 	
 ---- Mk1 Passive Camouflage Module ----
   *Cloak Does Not Fail When Attacking*
```
bleehu: Is there a better way/place to put this warning?

CharacterCreation/05_ArmourList.txt
```diff
@@ -656,9 +651,9 @@ Secondary Module Benefit: -12 Nanite Activation Cost,
...
 ==============================
 
-	=====   LEVEL 1 REQUIREMENT   =====
+=====   LEVEL 1 REQUIREMENT   =====
 	
-	---- Explosives Carrier ----
+---- Explosives Carrier ----
 Primary Module Cost:	375
 Explosives Skill Req:	10
 Prim Module Hard Pts: 	1
```
bleehu, line 626: I think it would work best for these description paragraphs would be to remove the "description:" and leave them as paragraphs. I think that it's pretty self-evident that these are descriptions. Also, they'll parse better.

bleehu (approved):
Great start! 
Does the Approximate Table of Contents help with the readability/usability of the document when looking at it from the plaintext/dev view? 
If it does, we should get the parser to ignore it. If it doesn't, then let's yank it and let the parser build an accurate table of contents.

CharacterCreation/05_ArmourList.txt
1sourcecontrol: What are we going to do about the rest of the items that use that terminology? Notably, every weapon with additional information uses something like "Description/Notes:"

Also, removing the item denotation ruins the consistency of the system; however, since the primary goal is to get CXDocs up and running, we can probably remove all the Description/Notes text from every item file.

bleehu: Let's hold off on that work until I can talk through that with ya. There might be a better long term solution.

1sourcecontrol:
Since we aren't doing a table of contents on every text page, I guess it'd be fine to remove the TOC from this. I made that for new players reading the txt files, long before the parser.

CharacterCreation/05_ArmourList.txt
1sourcecontrol: Are you saying I can't put in text with multiple lines after a stat declaration? If I make it all one line, it will be really hard to read. Could you delimit the following lines based on tabs?

CharacterCreation/05_ArmourList.txt
1sourcecontrol: I'll put it in the desc

1sourcecontrol: Well, it looks pretty terrible... but Idk what to do with it so I made a single-point list.

CharacterCreation/05_ArmourList.txt
1sourcecontrol: alrighty

Issue #198, opened 2017-11-16
Armor Piercing Level and Armor Points


Turtlelord26:
"That attack beats your AP level so you take 60 damage to your AP."

It might be nice to rename one of these so the abbreviations are different. Every bit of clarity helps.

bleehu:
The compromise 1sourcontrol and I originally made on this was that AP level is APL; distinct from APs. That said, I garuntee this is going to confuse a new player. Especially one who hasn't played an RPG before who is also trying to wrap their head around how Intelligence is a Stat, but Will is a Save.

1sourcecontrol:
@Turtlelord26 the specific sentence you quoted, where did you find it?
It should probably say "That attacks beats your APL so you take 60 damage to your APs." then there isn't any ambiguity around which 'AP' is referenced.

*edit* I prefer to use the first character of every word in an abbreviation, when possible. That way, there is more prompting for your memory. Versus calling Armor Piercing Level "AL" or something.

bleehu:
Except when a GM is talking to a player in-game. I don't think Turtle quoted that from the rules, I think he's using a hypothetical situation. I think it's plausible. Maybe even likely. @1sourcecontrol you're welcome to disagree and ask for playtesting if you don't think so.

1sourcecontrol:
>Except when a GM is talking to a player in-game. 

What are you 'except'ing? Are you responding to my first paragraph or the second? I'll assume the first (feel free to correct me later if i'm wrong).

I mean, if people use the correct terms then there isn't a problem, is there*? My goal here has been to make the rules as easy to adopt as possible -- which means I'm trying to design so the player doesn't won't have to memorize as many things.

*With a rudimentary understanding of the armor rules, there are several ways that the proper terms can be inferred by imperfect usage: APL is a single digit number, 90% of the time. APs are double and triple digit numbers, 100% of the time. A player may reduce your "Armor Piercing Level", but they will never do damage to an "Armor Piercing Level" that is a property of guns and armor. Damage can only be dealt to a type of hit-points: "Armor Points". With even one or two of these pieces among the rule reading, you can say, "hmm, ok, Turtlelord must be referring to Armor Points when he said "AP" because we already compared AP level".
I tried my best to come up with terms that are both intuitive to the mechanics and to the realism. That being said, I'm all for a different term if it makes sense because I'm as imperfect as the next cat.

Turtlelord26:
Confirm was a hypothetical GM statement.

[...]

PR #248, opened 2017-12-08
merge to dev,  for issue #198. Styleguide

1sourcecontrol:
#198. Also, doing some misc style guide capitalization fixes and grammar fixes.


bleehu (approved):
It looks like I need to add Hard Point, Shield Upkeep, Shield Strength and Armor Strength to the Glossary? It's ok if that's a separate PR.

1sourcecontrol:
@bleehu 
Hyess, that would be good. Good catch! I'll make an issue

PR #258, opened 2017-12-15
Stylized Items document

bleehu:
To be held in stasis until Bleehu and 1Source can catch up over the phone.


CharacterCreation/07_Items.txt
```diff
@@ -2,164 +2,178 @@
...
 Wall Charge ($840):
-Reduces a 1.5x1.5m section of wall to rubble, allowing entry. Clear areas 2m 
-surrounding the charge or take 70 explosive damage.
-Wall thicknesses:
-Wood - Up to 12in.
-concrete - up to 10in
-Reinforced Concrete - up to 8 in
-Steel plate - up to 2 in
+    Reduces a 1.5x1.5m section of wall to rubble, allowing entry. Clear areas 2m 
+surrounding the charge or take 70 explosive damage. Densities of material one
```
1sourcecontrol: I believe you are describing "thickness" below, not "Densities"?

_Side note:_ the price of a wall charge should honestly be like, $100. You can buy explosives that do 120 damage in a 3m radius for $220. 

https://github.com/bleehu/Compound_X/blob/3dbd18037679a57ad9f0ee08034888a68f8257d1/CharacterCreation/04_01_Melee_%26_Thrown_Weapons.txt#L266

CharacterCreation/07_Items.txt
```diff
@@ -2,164 +2,178 @@
...
 combatant's jaw bone. It uses the vibration of the jaw to send clear voice
 sounds even over cacophanous gunfire. A similar speaker is injected in the
 back of the cheek near the ear. Impossible to see with the naked eye without 
 invading the wearer's personal space. DC 140 to hack.
-
-Other misc. items are available for sale in the world. While you're playing be 
-on the lookout for shops or characters who might sell you more things. If you
-want to buy other starting equipment, ask your DM if something specific is 
-available and how much it might cost. Equipment that you start with generally
-depends on your character's fist mission.
```
1sourcecontrol, line 178: Do we want to add a section somewhere talking about this? The malleability of the game to allow additional unlisted items?

I am of the opinion that we should list out all the items normally available and that we should leave it as implicit that the DM can make up what they want. But I offer this question just to be thorough.

CharacterCreation/07_Items.txt
bleehu: Some scope creep, but ya, I thought that was a little pricey. I'll drop the price of Door breaches too.

CharacterCreation/07_Items.txt
bleehu: I'm really torn on whether to leave it open ended or not. Let's call it scope creep in terms of this particular PR, and open another one to tackle misc. items. I think I'm going to end up coming down on the side of not allowing unlisted items as well; that lets a dungeon creator know what items a character is likely to start with. But I think there's still room for debate. Since this is a mechanics question, the final decision will be yours, 1Source.

bleehu:
@Turtlelord26 can we merge this please? It will make https://54.200.202.155/docs/items look a lot better!

CharacterCreation/07_Items.txt
```diff
@@ -2,164 +2,178 @@
...
-For normal damage (50% or less) roll percents under 50. On pass, repair 30 + 
+Repair kit ($100):
+    For normal damage (50% or less) roll percents under 50. On pass, repair 30 + 
 percent rolled. For severe wounds (50% or more), roll percents under 50. On 
 pass, repair 30.
-If you have Skill in Repair, see the skill sheet for separate instructions.
 
-Nanite Pack: $100
-Restores 2d10 + 20 nanites.
+    If you have Skill in Repair, see the skill sheet for separate instructions.
```
Turtlelord26: Nonblocking inconsistency: Repair kits puts this line in a new paragraph but medkits do not.

CharacterCreation/07_Items.txt
```diff
@@ -2,164 +2,178 @@
...
 
-Det Cord ($230 per 2m of cord):
-Used to remove doors from hinges, act as ignition sources, enhance another 
-explosive, etc. Use 2m per door to be removed. .5m to wrap around a door 
-handle. 1m to wrap around a grenade. 4m to create a wall breaching pad (
-Explosives 50 DC)
+Wood:       Up to 12in.
+Concrete:   up to 10in
+Reinforced Concrete: up to 8 in
+Steel plate: up to 2 in
```
Turtlelord26: ~~Blocker~~: I am not certain what this will be read as. If we delete the preceding blank line will the parser interpret the DC lines as part of the preceding paragraph? (that would be good).

We could also turn DC lines into an unordered list.

CharacterCreation/07_Items.txt
```diff
@@ -2,164 +2,178 @@
...
+    Used to remove doors from hinges, act as ignition sources, enhance another 
+explosive, etc. Use 2 meters per door to be removed. One half meter to wrap 
+around a door handle. 1 meter to wrap around a grenade. 4 meters to create a 
+wall breaching pad (Explosives 50 DC).
 
+==============================
+Miscellaneous Items
+==============================
 
 Misc. Items and RP trinkets:
```
Turtlelord26: Blocker: redundant heading. Remove the topic one, keep the chapter one for consistency with the rest of the document.

CharacterCreation/07_Items.txt
```diff
@@ -2,164 +2,178 @@
...
-Trulight LED flashlight	$22:
-A metal black flashlight with bright LED bulbs. Illuminates up to 15m of bright light.
+Smart Phone	($180):
+    A touch screen, microphone and speaker in a plastic case. Reception may vary 
+by location.
 
-Smart Phone		$180:
-A touch screen, microphone and speaker in a plastic case. Reception may vary by location.
 DC to hack with skill points in hacking: 30
 DC to hack without skill points: 80
```
Turtlelord26, line 157: ~~Blocker~~: I am not certain what this will be read as. If we delete the preceding blank line will the parser interpret the DC lines as part of the preceding paragraph? (that would be good).

Turtlelord26 (changes requested):
Couple notes (non blocking):

There is inconsistent usage of blank lines after chapter headings (tools, wearables, etc). Style guide does not appear to require one way or the other.

Based on the descriptions of the different types of heading I would conclude that it is more appropriate to name the objects in section headers (or subsection headers, and make what are currently chapters into sections) instead of topic headers. That said, I definitely think it looks better (in github) with the topic syntax.

CharacterCreation/07_Items.txt
bleehu: This is set up as a stat block. We could instead do a table? If we wrap it into the paragraph, it will render them all on the same line, not the prettiest.

CharacterCreation/07_Items.txt
bleehu: wilco

CharacterCreation/07_Items.txt
bleehu: I don't think you really want the DC's to be in the paragraph, do you? It would look like this.

A touch screen, microphone and speaker in a plastic case. Reception may vary by location. DC to hack with skill points in hacking: 30 DC to hack without skill points: 80

CharacterCreation/07_Items.txt
bleehu: will change

CharacterCreation/07_Items.txt
Turtlelord26: I didn't see stat block when I was looking at the base styling doc and thus forgot they exist. mb.

CharacterCreation/07_Items.txt
Turtlelord26: Disregard this one too.

CharacterCreation/07_Items.txt
bleehu: sounds like my bad ;)

bleehu:
If the notes are non-blocking, can you approve the changes?

Turtlelord26:
Still waiting on taking out the repeat heading.

bleehu:
wilco

Turtlelord26 (approved):

PR #260, opened 2017-12-27
removed old pre-beta sheet

bleehu:
Not sure why this is still here.


Turtlelord26 (approved):
I certainly couldn't tell you.

PR #261, opened 2017-12-27
Stylized Basic Rules to conform with proposed table styles

bleehu:
We should probably hold this until we decide on table styles in the CX Design repo.


Basic Rules.txt
```diff
@@ -213,29 +213,31 @@ rolls are Perception and Dexterity. The Perception stat affects a ranged weapon'
...
++--------+-------------+	
+| 10-11  |  1.5	       |	
++--------+-------------+	
+| 12-13  |  2	       |	
++--------+-------------+	
+| 14     |  2.5	       |		
++--------+-------------+
+| 15     |  3	       |		
++--------+-------------+
+
```
Turtlelord26, line 240: Something strange happened to the trailing pipes here.

Turtlelord26 (changes requested):
Other than the one thing everything seems to be in order here.

Basic Rules.txt
bleehu: It be proper to my eye, sir?
it be proper to my eye

Turtlelord26: Iiiiinteresting. Here's what it looks like on my screen.
image

Turtlelord26: It's just an artifact of the changes view. I went to the raw file and it's perfectly normal there.
Disregard.

Turtlelord26 (approved):

PR #268, opened 2018-01-01
Skills doc draft from hulk

Turtlelord26:
For C2.0 playtest.


Turtlelord26:
Blee approves via Mark I over-the-shoulder review.

PR #226, opened 2017-11-29
Stylized Races document

bleehu:
To facilitate HTML display of rules in CXDocs and make reading easier.

PR #282, opened 2018-01-02
Addressing Racial Stylization Review

Turtlelord26:
Handled all blocking and nonblocking issues from my earlier review in #226, except for the inconsistent handling of racial trait style. I didn't want to unilaterally pick a way to do that, as I'm not on the team responsible for generating the traits.


bleehu (approved):
shipit!
Thanks for handling this!

PR #250, opened 2017-12-11
Glossary  and Basic Rules style fixes

bleehu:
Also added Free Action to Glossary as well. Also cleaned up mess not finished in PR #205

PR #284, opened 2018-01-02
Addressing Basic Rules Stylization Review

Turtlelord26:
Fixed all blockers and nonblockers from my recent review in #250, except tables.

### DIFF
--- a/Basic Rules.txt
+++ b/Basic Rules.txt
@@ -194,8 +194,7 @@ have 3 Movement Speed (from your Dexterity) and 4 Carry Ability (from your Forti
 Now, Strength is much higher than both of those, but Strength can't determine both 
 Movement Speed and Carry Ability so you have to decide whether you want:
 
-* 3 Movement Speed and 6 Carry Ability 
-* or 
+* 3 Movement Speed and 6 Carry Ability, or 
 * 6 Movement Speed and 4 Carry Ability 
 
 	Strength determining Movement Speed is likely to be the better choice, as 
@@ -211,15 +210,11 @@ shy or just have a terrible stutter.
 
 	But your raw base stat is rarely used in calculations during the game. To
 get the math to work out better, we usually us what is called a 'Stat Bonus'.
-Stat bonuses are used very often. They are calculated as so:
+Stat bonuses are used very often. They are calculated by one of the two equivalent 
+equations below (use whichever looks easier to you):
 
-	Stat Bonus = (4 * Stat) - 20 
-	
-	which is equal to 
-	
-	Stat Bonus = (Stat - 5) * 4
-	
-	if that's any easier.
+* Stat Bonus = (4 * Stat) - 20 
+* Stat Bonus = (Stat - 5) * 4
 
 	For instance, if you were trying to lift something heavy, your GM may tell 
 you to roll a 'Strength Check' and add your 'Strength Bonus'. To do this, you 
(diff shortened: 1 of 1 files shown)


Basic Rules.txt
```diff
@@ -544,14 +538,23 @@ Saving Throws
...
-- Will: defend against a mind affect (Ex: throw off hopelessness after crashing in an escape pod)
-- Shock: maintain composure after serious damage (Ex: stay conscious after being shot by a high caliber round)
-- Reflex: quickly react to an event (Ex: dodge falling debris, fire at an enemy if they leap from one piece of cover to another).
+
+- Will: defend against a mind affect (Ex: throw off hopelessness after crashing 
+	in an escape pod)
+- Shock: maintain composure after serious damage (Ex: stay conscious after being 
+	shot by a high caliber round)
+- Reflex: quickly react to an event (Ex: dodge falling debris, fire at an enemy 
+	if they leap from one piece of cover to another).
```
bleehu: Blocker: This change will confuse the parser. Before it would have rendered as a stat block (which on reflection might not have been the best way to do it) the new format will render as unknowns.

Basic Rules.txt
```diff
@@ -544,14 +538,23 @@ Saving Throws
...
-- Shock: 2*(Fort + Int - 6) Rational: You need to recognize that you are in shock, and resist its physical effects.
-- Reflex: 2*(Per + Dex - 6) Rational: In order to react quickly to an event you need to be able to see it coming.
+
+- Will: 2*(Cha + Int - 6) Rational: Normally wisdom would be applied, so Will in 
+	Compound X is descried as your internal voice motivating you to endure a 
+	mental hardship.
+- Shock: 2*(Fort + Int - 6) Rational: You need to recognize that you are in shock, 
+	and resist its physical effects.
+- Reflex: 2*(Per + Dex - 6) Rational: In order to react quickly to an event you 
+	need to be able to see it coming.
```
bleehu: Blocker: Same as above, Instead of a stat block, this will render as unknown.

Basic Rules.txt
```diff
@@ -625,7 +628,7 @@ back up if you have fallen.
 	The skill DC to stabilize a fallen comrade is 80. Medicine skill, INT mod, 
 and any bonuses from medical equipment such as medpacks may be added to that roll. 
 
-*Footnote: Some hazards can reduce you to zero health points and kill you 
+*Some hazards can reduce you to zero health points and kill you 
 instantly. Some examples are falling several stories, lava baptisms,
```
bleehu, line 625: This is very good! Thanks!

Basic Rules.txt
```diff
@@ -683,7 +686,7 @@ penalties if they attempt to use a skill that they don't have "Proficiency" in.
 A character has Proficiency in a skill when they have 10 or more skill points in 
 that skill.
 
-Defenders win ties for opposing rolls. If a character ties a DC they fail.
+	Defenders win ties for opposing rolls. If a character ties a DC they fail.
```
bleehu, line 682: Very good change! Good catch!

bleehu (changes requested):
A couple of small changes, then shipit!

Basic Rules.txt
Turtlelord26: Really? I expected the leading dash to take precedence and make it read as an unordered list. Will revert.

bleehu (approved):
shipit! Great work!

Issue #280, opened 2018-01-02
Gripes - Status Effects


Reveraine:
It would be nice to have effects in their own document, separate from the basic rules. To be referred to often and easily.

PR #286, opened 2018-01-03
Mh conditions section

bleehu:
To respond to gripe #280. If we are going to break the Basic Rules document up into multiple documents, we should also rename the files to keep an order as well.


Turtlelord26:
Shouldn't we delete the conditions from 01_01 if we're now listing them in 01_02?
Also why is the Glossary 03 when there is not an 02?

Reveraine:
I do think it would be nice to edit everything into the correct documents at once, but that's mostly so we make certain to not have overly redundant information.

bleehu:
Removed redundant section. What do you think now?

Reveraine:
Excellent. 
Although, I am also confused as to why Glossary is 03, when we don't have an 02. But if it's because you're planning on having a different 02 that's just not in yet, then this is good to go!

bleehu:
I do have the sneaking suspicion that once we start pulling very specific sections of the Basic rules out, we will end up having at least one more section. Not 100% sure what that is yet, but that's why I made that gap. 

Will blend-merge to fix conflicts.

PR #287, opened 2018-01-03
Added Roll Definitions to Glossary

Reveraine:
Average rolling was something I thought of while reworking races. I don't think it's too complex, but it will need some testing, I think. I'll put more thoughts for it in basic rules.

### DIFF
--- a/Glossary_of_Terms.txt
+++ b/Glossary_of_Terms.txt
@@ -89,6 +89,18 @@ Player Character at a time.
 
 RPG: 
     Role-Playing Game, such as Dungeons and Dragons or Compound X!
+    
+Roll - Standard:
+    Roll 1d100.
+    
+Roll - Advantage:
+    Roll twice. Take the highest of the two rolls. 
+    
+Roll - Average:
+    Roll 2d10s. Add together and divide by 2. 
+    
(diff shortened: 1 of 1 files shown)


Glossary_of_Terms.txt
```diff
@@ -89,6 +89,18 @@ Player Character at a time.
 
 RPG: 
     Role-Playing Game, such as Dungeons and Dragons or Compound X!
+    
+Roll - Standard:
+    Roll 1d100.
+    
+Roll - Advantage:
+    Roll twice. Take the highest of the two rolls.
```
bleehu: Should we add a "See the Primary Game Mechanic section of the Basic Rules" note in the description of each of these rolls? Not a blocking issue.

bleehu (approved):
One question, not blocking. Ship it if you'd like!

Rev, Way to go making your first few pulls! Just a reminder since this is your first couple of times that you get to push the big green merge button-of-win! Don't forget to delete the branch after it's been merged to keep the list of branches that are in use short. Way to go!!!

Glossary_of_Terms.txt
Reveraine: That could be nice. I can do that.

PR #292, opened 2018-01-04
Mh update engineer

bleehu:
Unhooked Engineer from the Nanite pool. Still need to do a balance regression using Excell to check the sanity of these quick numbers.


CharacterCreation/02_Classes.txt
```diff
@@ -188,17 +187,32 @@ you're not busy fixing what other people broke, you get to work on making new
 stuff. Which some knucklehead will probably, eventually, break too.
 
 Class Primary Attribute: Intelligence
+
+Class Feats: 
+=======
```
Turtlelord26: I'm not sure why 190-192 were added, but as they are redundant and style noncompliant I must ask that they return to the nothingness from whence they came.

CharacterCreation/Class-Specific Documentation/Engineer Processes.txt
```diff
@@ -1,20 +1,48 @@
...
+the Engineer uses a resource called Material. An engineer's extruder has a 
+built-in recycler which can be used to salvage usable material from a variety of 
+sources. Often, recycling spent, unused or damaged parts is part of the standard 
+procedure of a given process. Material accrued on mission persists between 
+encounters and over small rests.
+
+Munitions:
+    Munitions are used in the engineer's extruder to print more ... active 
+components. Due to regulations munitions extrusion material is not retained 
+between combat encounters.
```
Turtlelord26, line 26: We need a better flavor justification for enforced munitions dumping because outlaw pirate engineers won't care about ECSC regulations.
This is a blocker under notional new dev branch rules.

CharacterCreation/Class-Specific Documentation/Engineer Processes.txt
```diff
@@ -1,20 +1,48 @@
...
-nearest network mote or the engineer that deployed it.
+    Any repair process that an engineer uses on a cybernetic ally (even if that 
+process doesn't specify that it can only be used on cybernetic allies) leaves a 
+"network mote" on that ally. When calculating roaming range of drones built in 
+the build tree, the range of the drone is taken from the nearest network mote or 
+the engineer that deployed it.
 
-    Unlocking a Repair tree skill requires having unlocked the skill below it on 
+    Unlocking a Process requires having unlocked the Process below it on Skill 
 the tree. Please see the Engineer Skill Tree - Repair.png for dependencies.
```
Turtlelord26: "Process below it on Skill the tree"
I think maybe the Skill in there shouldn't be.

Also, we should call it a process tree instead of a skill tree now.

CharacterCreation/Class-Specific Documentation/Engineer Processes.txt
```diff
@@ -1,20 +1,48 @@
...
 
+=================================
+=====		Repair Tree		=====
+=================================
```
Turtlelord26: If this will parse "=====\t\tRepair Tree	\t\t=====" as a chapter heading then this is not a blocker, but I question the necessity. Of note, the Build tree chapter heading below is much more normal, so this is an inconsistency.

Turtlelord26 (changes requested):
I like lots of resource pools but I have been told that this is not necessarily true of the average player. However I suppose we can keep this for the moment, because when the tech revolution is finally implemented we can give the Engineer Hardlight construction, which could neatly collapse your energy and material pools into the same resource.

CharacterCreation/02_Classes.txt
bleehu: Will nuke

CharacterCreation/Class-Specific Documentation/Engineer Processes.txt
bleehu: I got an idea on that...

CharacterCreation/Class-Specific Documentation/Engineer Processes.txt
bleehu: nope, you're correct. That's totally non-compliant. Nuking it.

CharacterCreation/02_Classes.txt
```diff
@@ -187,11 +187,10 @@ you're not busy fixing what other people broke, you get to work on making new
 stuff. Which some knucklehead will probably, eventually, break too.
 
 Class Primary Attribute: Intelligence
-
-Class Feats: 
-=======
 Starting Wealth: $2580 + 100*Luck
 
+== Class Feats: ==
+
```
Turtlelord26: This still doesn't go here. The class feats subsection heading exists in its proper spot at line 227, consistent with all other classes. The class passives are not class feats.

CharacterCreation/Class-Specific Documentation/Engineer Processes.txt
```diff
@@ -32,12 +32,12 @@ process doesn't specify that it can only be used on cybernetic allies) leaves a
 the build tree, the range of the drone is taken from the nearest network mote or 
 the engineer that deployed it.
 
-    Unlocking a Process requires having unlocked the Process below it on Skill 
-the tree. Please see the Engineer Skill Tree - Repair.png for dependencies.
+    Unlocking a Process requires having unlocked the Process below it on Process
```
Turtlelord26: on ^the^ Process Tree.

Turtlelord26 (changes requested):

Turtlelord26 (approved):

Issue #290, opened 2018-01-03
Weakness To Poison Mechanics


Reveraine:
https://github.com/bleehu/Compound_X/blob/08c546897515dab2a3543e1eafc554be727aba25/CharacterCreation/01_03_Races_Tundari#L44

Weakness to Poison/Contagions needs to be added here by a balancer.

Turtlelord26:
I could see doing any of 
1. Roll poison and contagion related saves with disadvantage
2. Take a flat penalty to such saves
3. Take a penalty to initial infection saves but not to subsequent recovery saves.

My personal favorite of these at the moment is combining 1 and 3. I particularly like 3 because it communicates the feeling and experience of increased poisonability but doesn't necessarily apply the unfun ugh-I'm-poisoned-forever feels. May also result in easier balance of poisons.

What ideas does everyone else have?

bleehu:
I concur with Turtle's analysis. I think you're developing a good instinct on what is fun and what isn't. =)

PR #293, opened 2018-01-05
Added poison weakness rule.

Turtlelord26:
re issue #290


bleehu (approved):
Non-blocking: please capitolize Disadvantage to indicate it is a game term as opposed to a commentary on the socio-economic class of the roll. 

Great work, Turtle! Simple and effective. Shipit!

PR #306, opened 2018-01-19
stylized heavy weapons

bleehu:
'cause they don't parse well right now


Turtlelord26 (approved):
Line 323 vs lines 346 and 362 are styled inconsistently. All three are notes on nonstandard suppression behavior, yet one is a footnote and the other two are general notes. Propose that all three should be paragraph form general notes (and possibly this requires that the *s on the suppression lines in the relevant stat blocks be removed). However I won't mark it blocking since style is technically being followed.

Also there are missing blanks after the chapter headings. This is nonblocking as they are not explicitly required in style, but we have been doing them elsewhere and it does make the raw docs easier on the eye.

PR #308, opened 2018-01-20
made style changes to pistols to make prettier after parsing in cxdocs

bleehu:
looking forward to pretty pistols on cxdocs


Turtlelord26 (approved):
Nonblocker but missing semi-conventional blank lines after chapter headings.

Question: Will "Ranged miss chance" really look better as a paragraph than as a topic?

bleehu:
Answer: I'm not sure. If you think it would look better I can switch it.

Turtlelord26:
I think it would in github, but CXDocs display is more important and I haven't seen how it looks there.

PR #320, opened 2018-02-06
Remove lvl10 folder

1sourcecontrol:
The level 10 weapons, attachments, and armor are non-conforming to C2.0. They are nice references tho, for additional weapons and names. I didn't add any railguns, and I left out quite a few of the other weapons when I made C2.0.

Some of the items can be added to the current items list (with a lvl10 tag).


bleehu (approved):
Ya, we ought to dump this.
I think there's a lot of value in hiding high-level gear to be explored later. I'm sad we're not doing that. But if we do, this crufty stuff isn't what you should find.

1sourcecontrol:
If we ever release this game as a handbook, I think it'll be hard to hide later level weapons, if they are part of the release material.
I'm all for custom DM weapons and gear, or maybe mission specific loot that players don't have normal access to. I think the latter way would make the most sense.

Splitting the gear into multiple documents could be done, as Taylor did in the past. It leaves a sort of partial shroud. I suppose my reasons for wanting 1 document for all levels (per weapon type) are: less files to manage and all your equipment is easily compared.

PR #321, opened 2018-02-07
defining auto and burst fire rates

1sourcecontrol:
also clarified some text that defines "no crits on high fire rate weapons"


Turtlelord26 (approved):
👍 
Slight non-blocking concern, but the colloquiallism 'crit', which stands for "land a critical hit", might be unfamiliar to new players. It may behoove us to spell that out.

1sourcecontrol:
Spelled out, along with 1 other

Issue #316, opened 2018-02-06
Renaming "silenced" to be aesthetically compliant


1sourcecontrol:
https://github.com/bleehu/Compound_X/blob/973ff5fc803ba781bbc6c0171580a829b991bdc5/CharacterCreation/Class-Specific%20Documentation/Doctor%20Procedures.txt#L328

https://github.com/bleehu/Compound_X/blob/973ff5fc803ba781bbc6c0171580a829b991bdc5/CharacterCreation/Class-Specific%20Documentation/Doctor%20Procedures.txt#L330

While the term is common in fantasy games, I think it doesn't fit our lore.

Turtle suggested "stifled", I'm more for an explicit "Nanite Impairment" or "Nanite Disrupted/tion"

1sourcecontrol:
I think we should add a status affect if anyone thinks this might be used in other abilities. I could certainly see this getting used as a status effect on special bullets.

Turtlelord26:
This is probably blocked by the resource discussion, as it will be hard to name our silence effect without a solid foundation for what it's silencing.
(if someone disagrees and has progress to submit feel free to nix the blocked tag)

1sourcecontrol:
I looked all over the place - I thought we had an issue for the resource discussion- I'll go ahead and make one so we can track progress on github

1sourcecontrol:
#317

PR #327, opened 2018-02-10
Mh issue 316 disrupted

bleehu:
There was an accidental merge to dev, which was reverted. Which is why this took 3 commits instead of 1 clever one. 

@Reveraine says that "disrupted" implies an instant reaction to her, rather than an ongoing condition. If that resonates with other's interpritations, we'll have to find another word.

### DIFF
--- a/01_02_Conditions.txt
+++ b/01_02_Conditions.txt
@@ -18,12 +18,6 @@ Every 2 turns, the degree of the burn intensifies
     Your health has fallen below 0. See the "Death and Dying" section in the 
 Basic Rules Document.
 
-== Disrupted ==
-
-    You cannot use Nanite abilities, Doctor Procedures, Engineer Processes or 
-Hacker Exploits. Many similar abilities are also prone to Disruption. If they
-are not on this list, consult your GM. 
-
 == Frozen ==
 
     d10 damage per turn. lasts until a source of heat is applied to the wound, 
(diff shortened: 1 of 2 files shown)


bleehu:
Should fix Issue #316 once complete

Turtlelord26:
I would read the verb disrupt as an instant event, but the noun disrupted is a state or condition. The noun interrupted reads like an instant, but I wouldn't have read disrupted as instant if you hadn't brought it up here.
Silenced could also be interpreted as instant, but we don't because the term is ingrained as a condition in rpg settings.

1sourcecontrol:
>I would read the verb disrupt as an instant event, but the noun disrupted is a state or condition.

What @Turtlelord26  said.

### However, aren't we jumping the gun a little?

From what Turtle said in issue #316 (and I agree with):

>This is probably blocked by the resource discussion, as it will be hard to name our silence effect without a solid foundation for what it's silencing.

Although "disrupted" is generic enough to (probably) merge this PR without further ramifications... It'd be nice to know the aesthetics involved. If there has been a ruling on what the aesthetics are, can someone/ @Reveraine make an issue with the specifics?

bleehu:
Can we merge it as-is since it's compliant with current aesthetics? That way we have a slightly more stable current version? Then updated it or yank it following the aesthetic discussion?

Turtlelord26 (approved):
1Source says "probably [ok]", I say it's demonstrably better even if not ideal (which we have yet to determine), so 👍

1sourcecontrol (approved):
I guess I wanted to do this in a sequential fashion. Yes, merge.

What even is our current aesthetic? Nanites flow in the body?

PR #305, opened 2018-01-17
Mh version beta 0.0.3

bleehu:
Cutting Dev (with bumped version number) to RC. Time to send our hard work to testing!

PR #328, opened 2018-02-12
Mh version beta 0.0.4

bleehu:
This should be put on hold until the version beta 0.0.3 has been reviewed and merged.

PR #305  first.


Turtlelord26 (commented):
I question 
-sending the new player walkthrough to RC in its currently incomplete state, 
-that hackers are still called priests in classes but referred to as hackers in new skills,
-that we would merge new skills without deleting old skills, 
-and merging the new tundari document before the rest of the race reworks go live.
Other than that, things seem to be in order. Approval awaits only that the above points be spoken to.

bleehu:
Then let's remove what's not ready to be merged to RC. What do you think?

Turtlelord26 (approved):
:+1:

Issue #249, opened 2017-12-08
Min/Max penalties too strong?


1sourcecontrol:
When I recently made a new character, I considered putting a 1 in a stat. I wanted a more challenging character. However, when I looked over the list of stats, I felt that STR/PER/CHA penalties are too strong. If the penalties are too strong, hardly anyone will consider it an option because it isn't much fun to play the game as a vegetable --the penalties effectively bar usage. 
Here is the current list of debilitations:

- Strength: If you have 1 Strength, you are exceptionally weak and your max movement per 2 actions is 2m (unless your Dex is 1) and you cannot wear armor that offers more than 50% coverage.
- Perception: If you have 1 Perception, you are partially blind, only have 1 eye, or something similar. You cannot use the Search or Awareness skills and you have a -20 malus to Speech skill rolls. 
- Fortitude: If you have 1 Fortitude, you are exceptionally frail. You can't inflict lethal 	unarmed damage and your unarmed damage is halved. Also, when you deal melee damage, 	you inflict half of the damage on yourself.
- Charisma: If you have 1 Charisma, you have a debilitating speech impediment. You have an additional malus of -24 when using Speech, Deceive, and etc. skill. Under normal circumstances, you have to succed a DC 30 CHA skill check in order to communicate with a non-player character (NPC). In combat, vocal communication to anything costs 1 Action.
- Intelligence: If you have 1 intelligence, let's just say, you aren't particularly smart. You cannot attempt to use any skill that you do not have Proficiency in (see the 	"RP / Using Skills" section).
- Dexterity: If you have 1 Dexterity, you are especially clumsy or suffer from a debilitating	illness which affects your ability to move swiftly and precisely. You cannot learn any skills that use Dexterity as the contributing stat (for example, lockpicking).
- Luck: If you have 1 Luck, it goes without saying, you are very unlucky. At the start of a combat encounter, roll a d10 and receive a penalty from the Black Cat table, below.	

I recommend changing a few of these to the following:

- Strength: If you have 1 Strength, you are exceptionally weak. You cannot throw weapons more than 
	10m, and you cannot wear armor that offers more than 50% coverage.
- Perception: If you have 1 Perception, you are partially blind, only have 1 eye, or something similar. You cannot use the Search or Awareness skills and you have an additional -4 to your PER mod. 
[...]

PR #332, opened 2018-02-16
proposed a rewrite of penalty for having 1 strength

bleehu:
As per our discussion on Issue #249. Please note the movement impedance and weapon weight are baked into the weapons and armor documents respectively, and while mentioned here, were already defined.

### DIFF
--- a/01_01_Basic_Rules.txt
+++ b/01_01_Basic_Rules.txt
@@ -314,9 +314,13 @@ Strength (STR):
 how hard you can punch or hit with strength melee weapons. It might also allow 
 you to solve puzzles with the brute force approach.
 
-	If you have 1 Strength, you are exceptionally weak and your max movement per
-2 actions is 2m (unless your Dex is 1) and you cannot wear armor that offers more 
-than 50% coverage.
+	If you have 1 Strength, you are exceptionally weak. You are too weak to 
+effectively guard in melee, and thus your guard DC is always 0. In addition,
+you are too weak to control muzzle climb of any full-auto or burst fire. As such
+you can only fire weapons in semi auto or single fire. You are also limited to
+carrying the smallest of weapons and the lightest of armors. Please reference
+the strength requirement of each weapon and movement impedance of each piece of
+armor.
 
 Perception (PER): 
 	Your ability to see, hear, smell and sense the world around you, as well as


Turtlelord26 (approved):
This is more directly tied to the strength stat's core, 👍 

Though, as it looks even more penalizing, I'll defer explicit approval/not to 1source

edit: well that's :/ I meant to hit comment.

bleehu:
We'll wait on 1source, don't worry :)

1sourcecontrol (changes requested):
This goes far in the opposite direction of what I expected. 

> If you have 1 Strength, you are exceptionally weak. You are too weak to effectively guard in melee and thus your guard DC is always 0.

DEX determines melee guard, so now strength overlaps into DEX, trading the old movement speed penalty for a different penalty.

>  In addition, you are too weak to control muzzle climb of any full-auto or burst fire As such you can only fire weapons in semi auto or single fire. 

This is unnecessary since it's currently impossible to hold any weapon above strength requirement 3 (with weapon attachments) and all burst and automatic weapons start with SMGs (4 str req.) and above. If the DM were to grant them a weapon that has a lower strength requirement, they wouldn't be able to use it, why penalize the player?

> You are also limited to carrying the smallest of weapons and the lightest of armors. Please reference the strength requirement of each weapon and movement impedance of each piece of armor.

I'm not sure what 'smallest of weapons' means, unless, is it supposed to be colloquial?  
The player can have high FORT and DEX which would grant them high movement speed and high carry-strength. Is "lightest of armors" intended to be a further (yet unspecified) penalty?

1sourcecontrol:
I think Strength doesn't need many more penalties, once you have 1 STR. Here's why:

1. Implicitly, having 1 STR limits you from using some melee weapons (as some require strength, while others require dex)., And you can't use pistols unless the pistol has a STR requirement of 1 (they normally have a 2) or buy a "Light Profile Receiver (& Stock)" which costs $1000, a pretty hefty price at the start of the game.
I. Since many classes have access to more special weapons (marskman, hunter, specialist, soldier) having STR 1 dumps on class playability. 
II. Also, you won't be able to use certain Feats, now that there are more Feat stat restrictions.
2. You won't do as much melee damage because most weapons have some damage which is dependent on STR. It's not [STR or DEX] like it used to be, it's a [STR & DEX].  (1 exception being "Hand-to-Hand [Expert]" but STR gives a better gain than DEX). 
3. Your movement speed and carry weight will be determined by DEX and FORT, respectively. 

Although, I suspect the most obvious min-maxer's goal might be to max DEX and FORT and become the best tanky assassin out there. I'm going to sidetrack for a moment. I don't believe min-maxing should be punished _because it's min maxing_. I believe it should be balanced in a reasonable way that allows all players* to have fun. 

***Edit** Granting that the player is a non-sadistic, reasonable human being. P.S., please don't throw this 'granted statement' in my face. I think it's unreasonable to expect our players to be awful humans. 

To that end, I would like to compromise. I think it realistically (and mechanically, even if it has some overlap) fits to lower the max movement speed and inhibit thrown weapon range. One more possibility would be to double movement speed penalties from armor.. or in place of that, give the player a raw malus -5 movement speed penalty. I don't think that would work too well, though. As an assassin would have light armor, and if they min-max FORT and DEX, then they have 9+ Carry Strength. What about:
We could do a combination of forcing a max carry-strength as well as a starting movement-speed malus. Unfortunately, any way I can see to spin it, we'll be overlapping with DEX or FORT.
Proposals:
1. If you have 1 Strength, you are exceptionally weak. You cannot throw weapons more than
15m, Your maximum carry-strength is limited to 3, and you start with a -2m movement penalty.
2. If you have 1 Strength, you are exceptionally weak. You cannot throw weapons more than
15m and your max movement speed is 6 (not including ability modifiers). 

That way, the min-maxing assassin won't be able to achieve lightning running speeds like, more balanced assassins. And the throwing ability is for good measure (mainly affects grenade throwing ranges, which is tied to STR or DEX. But other throwing weapons are STR & DEX).

bleehu:
I don't feel good about merging this just yet. @Turtlelord26, I think that 1Source made some good points, but I'm not really sure how to proceed. I think maxing Dex as a way of making STR irrelevant is hokey and should be mechanically prevented, but I'm not really sure how. Do you see a way forward? If not, I'm going to just spitball another re-write until something sticks.

Turtlelord26:
PR on hold pending discussion that may render it obsolete.

bleehu:
@Turtlelord26, what were the results of that discussion?

Turtlelord26:
There hasn't been any discussion yet :(. 
See #397 

I'll toss in a few assignments to attract some attention.

bleehu:
@Turtlelord26, what do you think about aborting this pull until we see evidence of characters being made with a STR of 1?

Turtlelord26:
A good plan.

Turtlelord26:
Very stale, delete branch? @bleehu

PR #333, opened 2018-02-16
added Defuse and Jam DC to the Engineer's IED process

bleehu:
To address Issue #324


Turtlelord26:
Where it says proxy, do you mean proximity?

bleehu:
Indeed I did. Also I should have added some more detail on proximity detonation as well.

Turtlelord26 (approved):
[Stealth statement subject to revision in the far future]

Looks good 👍 

Though isn't that bit about concealing the IED the kind of rules-embedded tactical advice you've spoken against elsewhere?.

bleehu:
It borders on it, but offers some tactical constraints by stating which specific stealth roll is to be used.

PR #342, opened 2018-02-17
Humans - First pass

Reveraine:
Okay, hopefully I did this right this time xD

There are a few traits missing, still.
Feel free to add ideas if you have any =)

### DIFF
--- a/CharacterCreation/01_01_Races_Human
+++ b/CharacterCreation/01_01_Races_Human
@@ -1,0 +1,145 @@
+==============================
+Humans
+==============================
+
+===== Society =====
+
+Humans originated on the planet Earth in the Sol system. They have been exploring space 
+to some degree for almost 800 years, settling their own Terran Empire after the catastrophic 
+events that changed the Earth.
+In general, adult humans range from 5’ to 6’5”. They can be larger or smaller, but individuals
+of those types are generally rare. 
+They are an incredibly prolific and opportunistic species, having completely terraformed and 
+colonized over half a dozen planets and their respective moons. With many more Terran planets 
+on the rise, they already dwarf the small Opalite and Tundari civilizations.
+
+Humans are very culturally dissimilar to each other. Even more so than most other species, 
+most likely due to their rapid expansion. As such, it is impossible to pick out the cultural 
+traits that influence humans as a whole. Central government is important to the large groups of 
+humans, though smaller factions and settlements on the rim of their empire do not seem to care 
+about the rules and regulations of their governing powers.
+
+
+Types:
+* Biological
+* Mammalian
+* Warm-blooded 
+* Omnivorous 
+
+
+===== Language: Various ===== 
+
+Most humans speak what is known as Terran Common. In addition, humans have countless different
+languages that they use to communicate with others of their peers. The majority of humans 
+speak multiple languages, whether alien or domestic. Humans are very good at repeating 
+sounds that are initially unnatural to them. The language centers of their brains are highly 
+developed and well-used. 
+Terran Common, Eastern Common, Westlich, and Mandar are the most common human languages.
+
+[Humans gain 1 additional language and pick up new languages at twice the normal rate]
+
+
+===== Names ===== 
+
+Humans can be named in as many ways as there are Terran languages. Usually, these names 
+have some significant meaning to the parents, but that is not always the case. There is no 
+standard of naming that is generally agreed upon.
+
+== Human Name Examples ==
+
+* Matthew Kesler
+* Lupita Ugara
+* Tomoyo Usagi
+
+===== Available Traits =====
+
+Pitching Arm:
+All humans have the inherent bone and muscle structure necessary for extremely 
+accurate throwing. Not all humans practice this technique, but you are one who 
+has mastered this seemingly simple aspect of your anatomy.
+[Negates the aiming action necessary to throw accurately]
+
+Stubborn: 
+All Humans have a hardiness to them that cannot always be explained by logic.
+Some humans are just plain hard to kill, for no other apparent reason than they 
+just refuse to die. 
+[Roll Advantage on Death Saving Throws]
+
+
+==============================
+Human Sub-Species
+==============================
+
+===== Terrans =====
+
+Size: Medium
+Home System: Keplar
+
+Appearance: 
+Heavily resembling the historic humans, there is little variation from even 20th 
+century humans in this pool. The Terrans left Sol behind in the 24th Century, 
+forming a new colony in the Kepler system. 
+
+
+== Available Traits ==
+
+Very Normal: 
+Blending in has always come early to you. People tend to trust you. You seem to 
+belong no matter where you find yourself. 
+[Advantage on Persuade & Deceive ]
+
+Sixth Sense:
+Colloquially often known as the ‘gut’ instinct, sixth sense is that feeling that 
+many Terrans get when they feel something is off. There are different levels of 
+Sixth Sense, and many function in different ways. Some Terrans get goosebumps, 
+while others experience a hair-raising feeling; it can be very hard to get anything 
+past Humans with this sense tuned. 
+[Grants you an extra sense. DM’s discretion for application]
+
+
+
+===== Eastern Earthlings =====
+
+Size: Small/Medium
+Home System: Sol
+
+Inhabiting the areas of Earth east of equator, these humans have adapted to the humid,
+swampy heat of their half of the planet. Small and lean, these humans rarely reach over 
+5’ in height, and are much lighter than their Terran counterparts. As the new earth 
+conditions have stabilized, these humans have been beginning to rebuild some sense of 
+agriculture, but only recently have they ceased their nomadic patterns.
+
+== Available Traits ==
+
+Endurance: 
+Generations of Nomadic behavior have made moving long distances easy for those of 
+Eastern Earthling descent. 
+[First 2 Actions Sprinting are free of penalty. 1 Recovery Action required
+for every 2 Sprinting Actions]
+
+
+
+===== Western Earthlings =====
+
+Size: Large
+Home System: Sol
+
+Living as close to the equator as they dare, these humans are confined to the 
+western edge of the geostorm center. Some have spread further west, but most do 
+not like the absence of sunlight. As their climate tends to be cold to freezing, 
+these Earthlings are larger overall than their Terran counterparts.
+
+[Low-Light Vision]
+
+== Available Traits ==
+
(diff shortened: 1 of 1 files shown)


CharacterCreation/01_01_Races_Human
```diff
@@ -0,0 +1,145 @@
...
+* Lupita Ugara
+* Tomoyo Usagi
+
+===== Available Traits =====
+
+Pitching Arm:
+All humans have the inherent bone and muscle structure necessary for extremely 
+accurate throwing. Not all humans practice this technique, but you are one who 
+has mastered this seemingly simple aspect of your anatomy.
+[Negates the aiming action necessary to throw accurately]
```
1sourcecontrol, line 60: @bleehu When we've been playing CX, do we actually aim when throwing weapons?
I think I've had it both ways.

CharacterCreation/01_01_Races_Human
```diff
@@ -0,0 +1,145 @@
...
+
+Appearance: 
+Heavily resembling the historic humans, there is little variation from even 20th 
+century humans in this pool. The Terrans left Sol behind in the 24th Century, 
+forming a new colony in the Kepler system. 
+
+
+== Available Traits ==
+
+Very Normal:
```
1sourcecontrol, line 86: This has the same name as a feat. For clarity's sake I think we should either remove the feat or change one of the names. (preferably change one of the names)

CharacterCreation/01_01_Races_Human
```diff
@@ -0,0 +1,145 @@
...
+
+
+== Available Traits ==
+
+Very Normal: 
+Blending in has always come early to you. People tend to trust you. You seem to 
+belong no matter where you find yourself. 
+[Advantage on Persuade & Deceive ]
+
+Sixth Sense:
```
1sourcecontrol, line 91: I'll have to make an issue to remove Sixth Sense from the feats list, if we merge this.

CharacterCreation/01_01_Races_Human
```diff
@@ -0,0 +1,145 @@
...
+conditions have stabilized, these humans have been beginning to rebuild some sense of 
+agriculture, but only recently have they ceased their nomadic patterns.
+
+== Available Traits ==
+
+Endurance: 
+Generations of Nomadic behavior have made moving long distances easy for those of 
+Eastern Earthling descent. 
+[First 2 Actions Sprinting are free of penalty. 1 Recovery Action required
+for every 2 Sprinting Actions]
```
1sourcecontrol, line 118: Sprinting actions have been modified in C2.0, penalties don't stack indefinitely like they used to:
https://github.com/bleehu/Compound_X/blob/37f9e8d2537dc1844a7e710c06b2375f2572d812/01_01_Basic_Rules.txt#L859
So this reading of Endurance will be kind of break-even in terms of penalties. The longer you sprint, the worse it looks compare to C2.0 sprinting

CharacterCreation/01_01_Races_Human
```diff
@@ -0,0 +1,145 @@
...
+on the rise, they already dwarf the small Opalite and Tundari civilizations.
+
+Humans are very culturally dissimilar to each other. Even more so than most other species, 
+most likely due to their rapid expansion. As such, it is impossible to pick out the cultural 
+traits that influence humans as a whole. Central government is important to the large groups of 
+humans, though smaller factions and settlements on the rim of their empire do not seem to care 
+about the rules and regulations of their governing powers.
+
+
+Types:
```
1sourcecontrol, line 23: I may not have noticed before, but when I see "types" it makes me think... "something distinct that you only have 1 of." I.e., There is a biological type of human as well as a mammalian type of human, but they aren't the same thing. Not sure if that's just me, or wat.

1sourcecontrol (changes requested):
This is fantastic! I applaud your hard work!!

Couple of thoughts on the traits and the word "types".

FYI when most/all the races are done (or they are heading to R_C) I'll go through and try to make sure they are balanced. So it's not too much of a concern right now.

CharacterCreation/01_01_Races_Human
Turtlelord26: I've definitely always read it in the sense of qualities, as in humans have each of these listed things. Unfortunately it seems that the potential for confusion exists. Would "Qualities" be a better word? (Attributes is taken)

CharacterCreation/01_01_Races_Human
bleehu: I don't think people have been aiming with grenades especially. I didn't think we had a rule on it?

CharacterCreation/01_01_Races_Human
bleehu: It was super-clear to me. Maybe just 'cause it looks the same as the monster manuals from DnD, or the type from Pokemon. Types seems way more clear to me than Qualities.

CharacterCreation/01_01_Races_Human
Reveraine: In my experience, I don't know if it's written in the rules, but I'm pretty sure we've had a 'roll to hit what you're aiming at'. I could change it to reflect that?

CharacterCreation/01_01_Races_Human
Reveraine: This is effectively replacing the feat.

CharacterCreation/01_01_Races_Human
Reveraine: This is in place in order to designate how each species functions. It might seem redundant for humans because we're a little more familiar with humans than the alien species, but I have the same line in the Tundari Document.

CharacterCreation/01_01_Races_Human
Reveraine: Do you have a suggestion for a boon that will work better with C2.0?

CharacterCreation/01_01_Races_Human
Turtlelord26: Or even better, PR the deletion of sixth sense to this branch

CharacterCreation/01_01_Races_Human
```diff
@@ -0,0 +1,145 @@
...
+
+
+==============================
+Human Sub-Species
+==============================
+
+===== Terrans =====
+
+Size: Medium
+Home System: Keplar
```
Turtlelord26, line 76: Spelling inconsistency, should be Kepler if the intention is that the system be named after the mathematician. Nonblocking.

CharacterCreation/01_01_Races_Human
```diff
@@ -0,0 +1,145 @@
...
+[Grants you an extra sense. DM’s discretion for application]
+
+
+
+===== Eastern Earthlings =====
+
+Size: Small/Medium
+Home System: Sol
+
+Inhabiting the areas of Earth east of equator, these humans have adapted to the humid,
```
Turtlelord26, line 106: "east of [the] equator,"
Nonblocking.

CharacterCreation/01_01_Races_Human
```diff
@@ -0,0 +1,145 @@
...
+
+[Low-Light Vision]
+
+== Available Traits ==
+
+Quick Constriction: 
+Living in an area ravaged by lighting storms, these humans have developed rapid 
+pupil constriction and dilation. This increases their resistance to not only 
+lightning blasts, but to flash bangs. 
+[Effects from flash-bangs (and related implements) halved]
```
Turtlelord26, line 140: Flash-bang is hyphenated elsewhere in the rules. 140 is consistent but 139 has a nonhyphenated spelling.
Nonblocking.

Turtlelord26 (approved):
Some grammar comments, all nonblocking.

Overall this looks excellent (granting the stated need for a few more traits) and you have definitely piqued my curiosity about the history you're ~headcanoning~ writing.

Might be strong but what about a general trait similar to DnD's diehard, where the human can stay active for one or two rounds after being reduced to zero HP (but only gets 2 actions while at 0 and makes DSTs as normal)?

CharacterCreation/01_01_Races_Human
```diff
@@ -0,0 +1,145 @@
...
+Humans originated on the planet Earth in the Sol system. They have been exploring space 
+to some degree for almost 800 years, settling their own Terran Empire after the catastrophic 
+events that changed the Earth.
+In general, adult humans range from 5’ to 6’5”. They can be larger or smaller, but individuals
+of those types are generally rare. 
+They are an incredibly prolific and opportunistic species, having completely terraformed and 
+colonized over half a dozen planets and their respective moons. With many more Terran planets 
+on the rise, they already dwarf the small Opalite and Tundari civilizations.
+
+Humans are very culturally dissimilar to each other. Even more so than most other species,
```
bleehu, line 16: Nonblocking: We'll need to indent the paragraphs to get them to display well with the parser.

CharacterCreation/01_01_Races_Human
```diff
@@ -0,0 +1,145 @@
+==============================
+Humans
+==============================
+
+===== Society =====
+
+Humans originated on the planet Earth in the Sol system. They have been exploring space 
+to some degree for almost 800 years, settling their own Terran Empire after the catastrophic 
+events that changed the Earth.
+In general, adult humans range from 5’ to 6’5”. They can be larger or smaller, but individuals
```
bleehu, line 10: nonblocking: should adult sizes should be in the anatomy section?

CharacterCreation/01_01_Races_Human
```diff
@@ -0,0 +1,145 @@
...
+Types:
+* Biological
+* Mammalian
+* Warm-blooded 
+* Omnivorous 
+
+
+===== Language: Various ===== 
+
+Most humans speak what is known as Terran Common. In addition, humans have countless different
```
bleehu, line 32: Huzzah! We know what language people are speaking now! :throws confetti:

CharacterCreation/01_01_Races_Human
```diff
@@ -0,0 +1,145 @@
...
+Quick Constriction: 
+Living in an area ravaged by lighting storms, these humans have developed rapid 
+pupil constriction and dilation. This increases their resistance to not only 
+lightning blasts, but to flash bangs. 
+[Effects from flash-bangs (and related implements) halved]
+
+Frost-Free:
+The hardy life lived by your people has encouraged an innate resistance to cold
+[Take half-damage from cold. Roll advantage on any adverse effects related to 
+cold (sluggishness, frostbite, etc)]
```
bleehu, line 145: I really like this as a species feat :+1: :)

bleehu (approved):
Great work! shipit!

Issue #322, opened 2018-02-08
Polish Regression Override Engineering Process


bleehu:
firing an extra bullet is a little out of aesthetic. I need to adjust it.

PR #347, opened 2018-03-02
Polish up Regression Override

bleehu:
Updated Engineer Proccess "Regression Override" to aim more quickly rather than firing extra bullets. Should address Issue #322


Turtlelord26 (approved):
Issue successfully addressed without side effects. Shipit

PR #354, opened 2018-03-02
Mh version beta 0.0.5

bleehu:
We've almost caught up! 
Doing Version bumps more frequently should make the reviews much shorter and easier.


Turtlelord26 (approved):
Make it so

rosebrallier (commented):
[This is me struggling with the review UI. I figured it out.]

Issue #353, opened 2018-03-02
Explicitly type the damage types of engineer processes


bleehu:
Since some armors and characters have specific interactions with Explosive/Ballistic/Laser damage, I need to be explicit about what type of damage is being dealt at all times.
Adding Fire damage to the list of damage types in Basic Rules may be a good idea as well.

PR #358, opened 2018-03-07
Mh issue 353 engi dmg type

bleehu:
The most dramatic change here isn't that I added very specific damage types (most of which could have been easily and correctly inferred) to the Engineer Processes, but that in doing so, I felt compelled to add the damage types to the basic rules explicitly as well. 

I also added a new damage type, which I felt was necessary to capture the flavor of what the Engineer's processes were doing. Internal Damage. Which is very much subject to comment/review/dismissal.

### DIFF
--- a/01_01_Basic_Rules.txt
+++ b/01_01_Basic_Rules.txt
@@ -1215,6 +1215,38 @@ concussive damage (called blunt damage) that ignores armor but you can only deal
 1/2 damage, up to a max of 25 damage. Max of 35 damage if you are using a melee 
 weapon (like a club, sword, or even hitting them with your gun).
 
+Electric Damage:
+	Some damage is delivered to characters with a high current and voltage. This
+can lead to burns, organ failure, fried circuits and chaotic effects on the 
+nervous or control systems of a given character. This damage does not have any
+special effect on armor, but certain characters are more prone to electric 
+damage than others, while some characters are more resistant to it.
+
+Fire Damage:
+	Some damage is inflicted with conflagrations or general combustion. While
+this damage usually leaves an ongoing burning effect, this damage type does not
+imply that simply by being Fire Damage. Please read a given attack or event 
+carefully to see if you are simply burned once, or have an ongoing effect as
+well. Some characters are more resistant to fire than others. Similarly, some
+armor in the game helps mitigate fire damage more effectively than others. 
+
+Acid Damage:
+	Some damage occurs when a substance of a significantly different PH balance
+comes in contact with your player and chemically reacts in an adverse manner.
+Such cases are usually painful and smell bad. 
+	While taking acid damage doesn't always mean that you have an ongoing 
+effect, read the source of the damage carefully to see if in addition to the 
+one-time damage effect, you also have an ongoing effect to deal with as well. 
+Acid damage can affect armor adversely as well, but that is not implied simply 
+by an attack being labeled with acid damage. Such effects will be listed 
+separately in the attack, trap, etc.
+
+Internal Damage:
+	Internal damage can be sustained from things like parasites, malicious
+microbots, or simply pushing your character's physical body past its breaking
+point. Unless otherwise stated, internal damage always deals full damage and
+ignores armor. 
+
 ===== Protection Types =====
 
 Armor:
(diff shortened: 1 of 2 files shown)


bleehu:
This should address Issue #353. Mentioning it purely for the git tracking.

Turtlelord26 (changes requested):
I like damage types, but I also want to avoid damage type creep. To that end, I want to see each damage type we make have a mechanical and/or flavorful (both preferred) reason for existing.

My analysis by item:

Electric: An excellent type. It's mechanical and flavor purposes are the same: to specifically threaten (semi)mechanical characters and other electronic systems.

Fire: Good, but maybe can improve. A necessary type to have to describe certain environmental effects, and helps us give resistance to that class of ow to particular races. Also flamethrowers.
However, since we don't seem to be baking anything special into the damage type intrinsically, I would like us to consider rebranding it to Thermal Damage, and packing extreme cold effects under the same umbrella. If we're remanding burning effects to individual weapon descriptions, we can do the same for numbness inducement, frostbite, etc, and heat resistant equipment is also usually effective against cold environments too (with the exception of truly extreme conditions).

Acid: Excellent. The flavor of pH damage hard to capture with another type, and this type mechanically supports specifically attacking many items, objects, and structures. As noted here, shares a potential similarity with Fire/Thermal by inducing DoTs, but this is acceptable because, at least flavorly, the DoTs will require very different treatment to remove.

Internal: Seems like a handy catch-all for damage that bypasses all defenses. Here's a throwaway concern: Sith Lightning in SWtoR deals internal damage. Electrical surges are modelable by an internal damage type. This is only saved from redundancy by our Electrical damage having special race-based effects, which we would not want to manually include on every electrical effect in the game.
Here's the real concern I have. If Internal's mechanical niche is that it bypasses defenses, then Internal and Explosive are badly redundant. One of them needs to go, and I think I like Internal more, since it can describe more things. My first instinct for modifying frag grenades, rocket launchers, etc. to account for losing their damage type would be to split the damage into a small chunk of Internal or Blunt (representing the shockwave) and a large chunk of Ballistic (the shrapnel), which could be blocked by armor. My logic is that I'm not confident that shrapnel and buckshot are different enough to warrant a different damage type.
Intended as discussion points.

Sidenote: I'm really not sure about blunt, and would like to see it also folded into Internal. That might even be within the scope of this PR.

Oh, and adding the damage types to the Engineer processes is good and helpful. I agree with your selections.

Merge blocked by discussion points under Fire and Internal, not by specific change requests at this time.

bleehu:
Serves me right for scope creep. The change I was asked to make is blocked by the changes I wasn't. And half of these would have been let through if I'd broken 'em up into small changes. Still catching up on reading.

bleehu:
If we lose either explosive or internal damage, what happens to the EOD heavy armor that reduces nearby explosive damage, but shouldn't reduce hacked medical microbots that feast upon thy innards?

bleehu:
Similarly, if we collapse fire damage and ice damage into thermal damage, how do we model thermal damage to electronic targets? Machines fare quite well in the cold, but tend to be inhibited by their resisters melting.

Turtlelord26:
Re EOD heavy armor: my point becomes less applicable if explosive damage does not in fact bypass defenses. If we intend to have especially shrapnel-resistant armors then we probably shouldn't collapse explosive.

What do you think about folding blunt into internal though?


Re Thermal: I'm not at this time anticipating any Cold damage weapons (even if we did have a cold grenade, the temperature shock happens fast enough I think we can justify the damage to mechanicals). So the primary instances where you'd see Thermal (cold) damage would be environmental. In those cases, we would model the different tolerance of robots to cold by specifying that they don't take damage until much lower temperatures than the average humanoid starts hurting at.

bleehu:
I'm willing to entertain the idea of folding fire damage and cold damage into Thermal Damage; you've made a pretty good argument. But I'm gonna call 'scope creep' because if we want to change the places where fire damage already exists (including several dungeon repos) that's going to take a whole 'nother round of work.

bleehu:
added Issue #361 to track the work.

Turtlelord26:
Added #362 for the other point to allow merge of current work.

Turtlelord26 (approved):

Issue #350, opened 2018-03-02
Buff Refractory Spray


bleehu:
The Engineering Process "Refractory Spray" might be a little underpowered with Combat 2.0.

Quoth 1Source: 
"Also suggest buffing this damage protection to 50 points. Lasers and plasma are more common-place in the lower levels now... so this ability may be a little underpowered (especially comparing to Ceramic Catalyst.)"

PR #359, opened 2018-03-07
buffed Refractory Spray 

bleehu:
as per Issue #350.

The damage disparity comes from each of the laser weapons that 1source made have lower damages on average as a trade-off for ignoring armor. 

Refractory spray is a 6th level Proceedure. It's first version ignores 80% of the next hit of a laser or plasma weapon, and then at 10th level is re-balanced to do the same for tougher weapons.


Turtlelord26 (approved):
I'll buy the flavor of the damage reduction discrepancy, though I am slightly concerned that it's twice as much to remember. That said, the ability is probably to the point where checking the memory issues (and numbers, of course) belongs properly to Testing.

PR #360, opened 2018-03-07
MH fishing pole

bleehu:
Here's that fishing pole nobody was asking for, but I thought might be cool. 

CC'd Masterdiekillsu since he has a character that might have seen one in living memory.


Turtlelord26 (approved):
So how many credits will bait cost?

bleehu:
Make your own with survival check?

PR #379, opened 2018-03-19
added Juggler Feat

bleehu:
It bothers me that _something_ like this doesn't exist yet. We can pick this to pieces if we want; I'm not super emotionally invested in it. But we need _some_ feat for throwing knives and shurikens at least, otherwise I think they're way underpowered.


Turtlelord26:
DEX prereq seems low?

bleehu:
Low dex to attempt, but at that level you have an almost 50% chance to drop all of your grenades at your feet rather than lobbing them at the enemy. In order for the feat to be useful, rather than just funny, it is highly recommended that a character have a higher DEX.

Turtlelord26 (approved):
Ah, I hadn't actually followed that math. :+1: Testable feat.

Issue #373, opened 2018-03-15
Gripe: Large magazines should take up more slots than small


rosebrallier:
Machine gun and heavy weapons mags should take up two standard mag slots on armor due to their large size.

PR #380, opened 2018-03-19
added 2 slot requirement for large munitions

bleehu:
Should fix Issue #373. Intentionally left antimaterial single rod shots as a single mag, as well as a few of the laser guns, which I think would have smaller batteries.

### DIFF
--- a/CharacterCreation/04_06_MachineGuns_HeavyWeps_&_RocketLaunchers.txt
+++ b/CharacterCreation/04_06_MachineGuns_HeavyWeps_&_RocketLaunchers.txt
@@ -9,7 +9,9 @@ without 2x+ magnification.
 MACHINE GUNS
 ==============================
     For sustained automatic fire, machine-guns are the first choice.
-Minimum Strength 6. Reload time is 3 Actions. Jam range 2%.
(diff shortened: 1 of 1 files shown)


Turtlelord26 (changes requested):
I would like the mag size note under machine guns to be copied to the Heavy weapons umbrella description.

Re. Anvil Arms XM-500 Minigun, why does this one not have large magazines?

bleehu:
Can do, will do!

Uh, it definitely does. If I recall, that was the specific weapon that caused the gripe. https://github.com/bleehu/Compound_X/blob/83fc1c76289e7e9035b1ec1b96b8a344a0b0ae89/CharacterCreation/04_06_MachineGuns_HeavyWeps_%26_RocketLaunchers.txt#L225

bleehu:
Whoops! That's the microgun, not the minigun...

bleehu:
https://github.com/bleehu/Compound_X/blob/83fc1c76289e7e9035b1ec1b96b8a344a0b0ae89/CharacterCreation/04_06_MachineGuns_HeavyWeps_%26_RocketLaunchers.txt#L317

Turtlelord26:
Touchè

Turtlelord26 (approved):

PR #378, opened 2018-03-19
added hacking DC to ASPRO-S

bleehu:
Low priority; I don't think anyone uses an ASPRO anyway.

PR #381, opened 2018-03-19
Intercepting projectiles is not defusing.

Turtlelord26:
re #378

### DIFF
--- a/CharacterCreation/Class-Specific Documentation/Engineer Processes.txt
+++ b/CharacterCreation/Class-Specific Documentation/Engineer Processes.txt
@@ -977,7 +977,7 @@ Health:				50
 Size:				1m x 1m x 3m
 
     You assemble an active countermeasure system that eliminates incoming 
-explosive ordinance. When active, this system destroys any incoming grenades, 
+explosive ordnance. When active, this system destroys any incoming grenades, 
 missiles, rockets or thrown weapons that it has line of sight to from the top of 
 its 3m mast within a 6m radius. Any incoming munitions are destroyed before they 
 can detonate. Cannot stop incoming bullets, lasers, or plasma. Will stop some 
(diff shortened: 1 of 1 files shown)


bleehu (changes requested):
That seems _really_ nitpicky. Strictly speaking, using incompressible liquid to render an explosive unable to detonate is defusing it. https://en.wikipedia.org/wiki/Bomb_disposal#Projected_water_disruptors Simply intercepting a grenade would mean that it still goes off at the point of interception. The principle use of the ASPRO-S isn't just to keep the grenades away, but to stop them from going off. https://en.wikipedia.org/wiki/Trophy_(countermeasure).

Turtlelord26:
I have to concede the point about interception, but defuse is still wrong due to common usage issues. The technically correct "render less harmful" definition mutating into nonexplosive destruction just doesn't happen in practice, by comparison with the "reach into the bomb's guts (physically, electronically, whatever) and pull out the bit that makes it go boom".
At least, as far as I have ever seen.

I have tried a different approach that uses neither term.

And also fixed the issue where the ASPRO stopped incoming legislation instead of bombs.

bleehu:
How about "neutralizers" Which is what the Israelis call the projectiles that the TROPHY system uses?

bleehu:
My concern is that when I hear "ammunition" I think "bullets". and this doesn't fire bullets, it shoots jets of basically hydraulic fluid.

bleehu:
ah, nevermind. It's not worth quibbling over. If you like ammunition, let's just go with that.

bleehu (approved):

Turtlelord26:
I think I've managed to sidestep the issue of what it's firing entirely, which may be for the best. If we want to specify what it's shooting, we should do so higher in the description, maybe tacked on to the "eliminates incoming explosive ordnance" first sentence.

bleehu (approved):
shipit!

Issue #351, opened 2018-03-02
Define Engineer IO Ports


bleehu:
They are mentioned, and intuitive to use, but not really explicitly explained.

Also, add them to the Glossary.

PR #388, opened 2018-03-22
added detailed description of IO Ports as per Issue #351

bleehu:
I'll take a review from basically anyone on this. 

the title didn't quite make the automatic link to Issue #351 so I added it here.


rosebrallier:
The description you have is clear and precise.
Nitpick: some players will want to know what, if anything, "IO" stands for.

bleehu:
That is a very good point! I can't believe I missed that opportunity!

Turtlelord26 (approved):
I concur with the above. This is functional, but would benefit from the explicit spelling-out of Input/Output re not-electrical-engineers.

PR #389, opened 2018-03-22
Mh version b 0.0.6

bleehu:
@Turtlelord26, you already reviewed almost all of these changes as their feature branch review. But please put a very close eye on this version bump review; I'm trying a new workflow using Git cherry-pick instead of git-merge with a bunch of reverts. I think I got it to work, but please double check it. 

Also, only like 6 updates, 'cause we're version bumping more often!


Turtlelord26 (approved):
Everything here looks good and familiar. Make it so.

PR #390, opened 2018-03-22
added IO ports to glossary and added actual definition of IO in IO Ports

bleehu:
Changes to be committed:
	modified:   ../../03_Glossary_of_Terms.txt
	modified:   Engineer Processes.txt

### DIFF
--- a/03_Glossary_of_Terms.txt
+++ b/03_Glossary_of_Terms.txt
@@ -83,6 +83,12 @@ Held Actions:
 to a trigger outside of your turn. See "Reactions & Held Actions" in the basic 
 rules document.
 
+IO Ports:
+    Short for Input/Output Ports. Engineer Class Characters can build Drones and
+some Buildings which have a number of IO Ports which can have Peripherals added
+to them. See "Character Creation/Class Specific Documents/Engineering 
+Processes.txt" specifically the "IO Ports" subsection for more information.
+
 Nanite Upkeep:
     Nanite Upkeep is the number of nanites a character must pay at the end of 
 their turn in order to stay cloaked during that round. For more information,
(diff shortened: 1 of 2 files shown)


CharacterCreation/Class-Specific Documentation/Engineer Processes.txt
```diff
@@ -520,6 +520,9 @@ saying this is that if a Drone has 3 IO Ports, you can't build 4 Peripherals
 onto it and have the Drone turn peripherals off or on as it wants to; every
 Peripheral must have an active IO Port or it simply falls off of the Drone. 
 
+    IO Ports are a real-world Computer/Electric Engineering term which is short
+for Input/Output Ports.
```
Turtlelord26, line 524: We're still going to lose people over Ports. It's not necessarily a problem, but I see the possibility for supplying learning and interest by adding something like " ... Input/Output Ports: physical connection points that allow two-way data transfer."

03_Glossary_of_Terms.txt
```diff
@@ -83,6 +83,12 @@ Held Actions:
 to a trigger outside of your turn. See "Reactions & Held Actions" in the basic 
 rules document.
 
+IO Ports:
+    Short for Input/Output Ports. Engineer Class Characters can build Drones and
+some Buildings which have a number of IO Ports which can have Peripherals added
```
Turtlelord26, line 88: Fairly nitpickish grammar, line 88 should read "some Buildings that have a number of IO Ports, which can have Peripherals added".
The first which is changed to a that, and the second which is preceded by a comma.

CharacterCreation/Class-Specific Documentation/Engineer Processes.txt
```diff
@@ -520,6 +520,9 @@ saying this is that if a Drone has 3 IO Ports, you can't build 4 Peripherals
 onto it and have the Drone turn peripherals off or on as it wants to; every
 Peripheral must have an active IO Port or it simply falls off of the Drone. 
 
+    IO Ports are a real-world Computer/Electric Engineering term which is short
```
Turtlelord26, line 523: Similarly nitpickish grammar, the 'which' in line 523 should be 'that'.

Turtlelord26 (approved):
Good work, can be shipped.
Some low-effort points noted that could make it even better.

Issue #130, opened 2017-07-12
Level 10 special ammo missing


bleehu:
We have no level 10 special ammo.

trowl223:
As a suggestion, Vorpel rounds and phasing rounds were somewhere in one of my ideas documents. 

- Vorpel rounds add a bonus to crit and extra damage on crit, something in the realm of -10% to each crit tier, +10 damage. Magazine cost * 2.5 or 3.
- Phasing rounds add +10% to hit, even when target is in cover. Magazine cost *1.5

bleehu:
What's a vorpel?

bleehu:
I think I found it! https://www.urbandictionary.com/define.php?term=vorpal

It would be nice to add Gluon and shredder rounds to level 10 weapon rounds.

1sourcecontrol:
Wat! I've never heard of a Vorpel. It's more famous than I had thought. 

Can you add a comment for what the Gluon and shredder round stats are? We could add them to the attachments doc (after a C2.0 balance check)

Turtlelord26:
Slightly more reputable source material for Vorpal, which is an adjective, not a noun.
http://www.dictionary.com/browse/vorpal?s=t

Obligatory aesthetic advisory.

bleehu:
==================================================
//	Ammo 
==================================================

	Baffled Ammo
--------------------------------------------------
Reduces noise of bullet travel to zero and reduces noise of impact to a whisper.
Zero RP cost, but bullets are made of an organic material that decays after two
missions.

+ stealth rolls to hide silent weapons fire automatically succeed. 
+ a bullet impact can be heard if a character is within 5m of the impact, 
	actively curious and makes a search check over DC 50.
- reduces range by 10 m
- reduces damage by 5

--------------------------------------------------

	Gluon Rounds
--------------------------------------------------
These rounds contain an incompressible, unshatterable Gluon matrix. These 
rounds are manufactured by a luxury weapons corporation and are not available
to the public or military. They occupy 3 RP per clip.

+ Pierces armor and shields, no damage reduction for piercing cover
- after the tenth shot, 5% chance to jam weapon

--------------------------------------------------

	Shredder Ammo
--------------------------------------------------
Rounds splinter into hyperkinetic shrapnel upon impact. No RP cost,
but Rounds loose hyperkenetic charge after 2 missions and then
act like hollowpoints.

- multiplies armor resistance of target by 2.5
+ deals an additional 40 damage on hits
[...]

PR #393, opened 2018-03-23
added Vorpal Rounds

bleehu:
Part 1 of 5 for Issue #130. Using MH lv 10 branch as a superbranch, will close the Issue when the proposed rounds are merged or rejected.

### DIFF
--- a/CharacterCreation/04_00_WeaponAttachments.txt
+++ b/CharacterCreation/04_00_WeaponAttachments.txt
@@ -457,14 +457,14 @@ armor or if it doesn't hit armor.
 == Vorpal Ammunition ($2.5 x mag cost) ==
     
     Vorpal Rounds carry a small nanite payload which damage soft, vital tissue
-or circuitry. Unfortunately, the nanites are over-equipped with microweaponry,
-but under-equipped with locomotive flagella. So they only really deal the extra
-damage if you put the round where it's gonna hurt.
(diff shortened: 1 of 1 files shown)


Turtlelord26 (changes requested):
1. I know you don't generally like to handle typography in initial PRs but please fix the spelling of Vorpal; it hurts to look at.

2. As we are aiming for a soft sci-fi, we need at least semi-plausible reasons for our things working as described. So what makes these bullets extra effective specifically when they hit particularly vulnerable bits of anatomy?

bleehu:
1. If you insist.

2. This was never explained to me. Scrap the round entirely?

Turtlelord26:
We can try posing the question to the team and see if anyone comes up with something half-plausible.

bleehu:
Is that still too hand-wavy?

Turtlelord26:
Definitely getting there. I think I would like a phrasing that implies that the lack of locomotion is the result of fundamental design constraints rather than a mistake (at least, I read over-/under-equipped as mistakes). Maybe a phrasing more like 'the density of nanoweaponry precludes powerful locomotive capability'.

{While you're in there, there's a damage that should be a damages in line 459}

bleehu:
How'zat?

Turtlelord26 (approved):
Still typoed :( but shipitable 👍

Turtlelord26:
😃

PR #394, opened 2018-03-23
added gluon rounds

bleehu:
Basically level 10 armor piercing rounds.

### DIFF
--- a/CharacterCreation/04_00_WeaponAttachments.txt
+++ b/CharacterCreation/04_00_WeaponAttachments.txt
@@ -431,6 +431,16 @@ Takes up twice the space in your magazine inventory as a standard magazine.
 Only applicable to box-fed automatic weapons [not belt-fed], SMGs, 
 Carbines, and shotguns.
 
+== Gluon Rounds ($2.5 x mag cost) ==
+    Gluon rounds use a web of inseparable gluon entangled particles to make a
+round that won't fragment no matter what it hits. That means that it delivers
+all of its energy to one point, which helps penetrate armor even better than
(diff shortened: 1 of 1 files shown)


bleehu:
Stats totally negotiable

Turtlelord26 (changes requested):
I'm mildly concerned, aesthetically, about absolutes like "inseparable" and "no matter what it hits". The mechanics seems reasonable.

It's not clear to me whether that last line is saying that gluon rounds bypass AS.

bleehu:
That's not actually Aesthetic. It's real-world science. I'm having a little trouble finding the specific formula, since I got the idea from a lecture at UW. But it turns out that the strong atomic force between to mated gluons increases geometrically the further apart they're pulled. With no known limit. https://en.wikipedia.org/wiki/Glueball. Now creating a glueball would take unrealistic amounts of energy, but significantly less than it would take for FTL travel.

Turtlelord26:
Strong force is breakable though. Yes, it does take more and more force to pull strongly interacting particles farther apart, but at some distance the strong force suddenly drops to ~zero~ negligible levels. Otherwise fission would not be possible.

Parts of my hightech document imply some control over the strong force, hence I'm not concerned about constructing them.

Also, more practically, re "no matter what it hits", if I shoot a glueball bullet at a glueball shield, which wins?

Turtlelord26:
Post discussion, now asking only for a clarity pass on the armor bypassing mechanic.

Turtlelord26 (approved):
Shipit

Query: I thought you wanted to try triple AS damage first?

PR #401, opened 2018-03-26
added Short Circuit rounds

bleehu:
Part 5 of 5 for Issue #130.


Turtlelord26 (approved):
Good, shipit.

Though, something to keep an eye on during testing: since the bonus and penalty are flat, miniguns with this ammo become absolutely terrifying to cybernetic characters.

Issue #209, opened 2017-11-22
Lanugages are a mess


bleehu:
So... Languages are a mess. They apparently are given out by your Race, but some are given "ECSC standard" some "Basic" and some "Emperial Common" [SIC]. Opalites don't even have it mentioned. (That last one is my fault.)

To resolve this ticket specifically, can we yank language as a hard-"coded" rule system as a whole and maybe re-implement it, maybe not? 

If/When we do re-implement language, maybe we can add a separate small rules document listing the available languages. Maybe we can add ways of learning languages as you level up? Maybe we can even have a semi-realistic sociological model wherein each species has more than one language and maybe even have language cross the Racial boundary, but fall more along political or geographic boundaries?

Reveraine:
While language and race makes sense, I think language and background make more sense to go together?  We can assume that each alien race has at least one language, but you are right, they need to be listed if they are a big part of immersion in the game.

1sourcecontrol:
I think we should either yank or homogenize our current implementation. Idk about the rest of everyone but I miss the DND aesthetic where I could speak Elvin and common, and it really came in handy sometimes if I was the only elvin speaking person. Or I could talk to another teammember without anyone understanding us.
But I think we could just homogenize the current implementation and give everybody common or Imperial common until a later date when we decide what to do with it. 

If each species has more than one language.. I would get headaches. Unless it's well segregated by each individual race in the race doc (like if elves were a subclass of humans, for instance) and everyone speaks common. And we'd run into problems where an entire party couldn't communicate if each character took a different language and didn't standardly have common.

Reveraine:
There was some talk of having a Universal Translator of some sort. This way, for the most part it would be an aesthetic choice. Unless a DM specifically wanted send their players through a 'You UT broke, Does anyone speak Fish?' scenario.

I agree with not making it too complex. I think a universal translator analog helps keep the realism alive without adding complication for new players. 

As far as for now, yanking them seems fine as long as we continue dialogue about it.

Reveraine:
Also!
Programming languages are incredibly relevant. For Example: Do nanites all speak C++, or can you program some with Python? What is droid X written in and do you have the equipment to hack that?
This should be an incredibly simple language system, if we decide to use it, but it's a thought I've had.

bleehu:
Let's continue the design discussion here, and I'll make another issue ticket to track the work of yanking the current language system.
[...]

PR #403, opened 2018-03-26
Universal Translator Item

bleehu:
This should close out Issue #209. Note that the translators are agnostic to the actual language mechanics, so they can be tested without having to be dependant on Rev completing all of the new race work.


Turtlelord26 (approved):
Hacking the translator is evil. 👍 

I would like the UT to come with a pair of cheap earbuds, but that could be pushed later.

bleehu:
Would it be cruel to make you pay for the earbuds separately and have you roll PER not to lose them?

Turtlelord26:
As cruel as it would be to charge 50000% of what it costs to make the things and design a 6-month lifespan into them deliberately.

Issue #231, opened 2017-11-29
New Player Guide Missing Section: Classes


Turtlelord26:
Pending class system rework.

bleehu:
I think the new class system is going to be delayed in starting, and will take long enough in production that we may as well handle this section for the current version. I think it's going to be very helpful for play over the next couple of months.

bleehu:
Assigned to Turtle, but being the director, Turtle can reassign as he sees fit.

PR #405, opened 2018-03-27
New Player Guide - Add Step 2: Class

Turtlelord26:
See #231 

While writing this up, I considered the cliche situation of new-player-dragged-to-the-table-because-the-group-needs-a-healer. I then decided to ignore it because Doctor is *not* one of our starting classes. We will want to consider adding off-healing subclasses during the class pass.


bleehu (approved):
shipit!

Issue #371, opened 2018-03-12
Feature Request: A Flamethrower Peripheral for Engineer Drones


Turtlelord26:
Someone more familiar with Engineer design history tell me if there's a particular reason we don't have them. There has been a request.

bleehu:
Why didn't I think of that? I'll add it to my queue!

PR #406, opened 2018-03-29
added Flamethrower peripheral

bleehu:
Should address Issue #371, Did not update the build chart yet, as working on graphics with my linux machine means using GIMP, which is soul-sucking. Will add later. I intend to add the flamethrower at about the same level in progression as the minigun.


Turtlelord26:
I imagined the drone flamethrower as smaller than the specialists'. What would you say to cutting the damage and decreasing the unit of fire (effectively increasing mag size)? Maybe damage to 40-45 and 1L per shot? Would decrease dps and/but reflect that it just can't spew gel as fast as the full-size version.

bleehu:
Definitely amenable to nerfing it. Discussion on specifics in slack?

Turtlelord26 (approved):
Let's test it!

PR #408, opened 2018-03-29
stylized carbines

bleehu:
I would also like to put the burst and auto fire into the range/accuracy table. But not sure how to denote the strength bonus.


Turtlelord26 (changes requested):
Expected deletions: lines 329-332, lines 397-400

Turtlelord26 (approved):
👍

Issue #165, opened 2017-10-25
Engineer Skill Tree Structural Concerns


Turtlelord26:
Most notably, I would like to point out that it's a little punishing as-is to make an engineer without the rest of the party being robots. The repair tree, which is the resource generator tree, has an enormous reliance on skills that can only target cybernetic allies. Below is a fully enumerated list of skills in the repair trees that do *not* require a cybernetic ally target (note that most of them are still gated behind a skill that does require such a target)

Restore Armor - Cyberphysical - gated behind 1 cyber only (Reboot)
Percussive Maintenance - Cyberphysical - gated behind 2 cyber only (Reboot & Data Scrub)

Duct Tape - Material
Weapon Tuning - Material
Vehicle Repair - Material
Engine Tuning - Material - Gated behind 1 cyber only (Micro-assembler)
Zero Friction Covalent - Material - Gated behind 1 cyber only (Micro-assembler)
Sup it Up - Material - Gated behind 1 cyber only (Micro-assembler)

Repulsor Recharge - Optronic
Repulsor Overcharge - Optronic - Gated behind 1 cyber only
Repulsor Projection - Optronic - Gated behind 1-2 cyber only
Refractory Spray - technically* - Optronic - Gated behind 1-2 cyber only
Tune Laser - Optronic - Gated behind 1-2 cyber only
Repulsor Hypercharge - Optronic - Gated behind 1-2 cyber only
Broadcast Repulsion - Optronic - Gated behind 1-2 cyber only
CS Lacer - Optronic - Gated behind 1-2 cyber only

I propose that we might consider a little rearrangement or allowable target adjustment to more fully separate "help your robot friends" and "everything else" as the former is a niche function.

Additionally, while I was combing through the trees, I found a few more miscellaneous concerns:

Colocation Processing/Dedicated Process - Cyberphysical - Have significantly different names in processes.txt and the tree diagram
Minesweep Pulse - Optronic - Is a node in the tree but has no description in Engineer Processes.txt
Gun Emplacement - Statics - has a description but no tree node
Snare - Statics - Has a description but no tree node
"??????" - Statics - :(

- [ ] Addressed cybernetic ally target structural concerns
- [x] Resolved name ambiguity in "Process" sklills
- [x] Added description for Minesweep Pulse or removed it altogether
- [x] Added tree node for Gun Emplacement or removed it altogether
- [x] Added tree node for Snare or removed it altogether
[...]

PR #409, opened 2018-03-30
Replaced old Engineering trees with new ones.

bleehu:
Added Colocation processing, Gun Emplacement and Snare.
This should handle many of the issues in #165


Turtlelord26 (approved):
Like 👍

PR #418, opened 2018-04-02
buffed gluon rounds 'cause they're pricey

bleehu:
I'm realizing now, I probably should've put up a kanban board and made level 10 ammo a project...


Turtlelord26 (approved):
👍

PR #419, opened 2018-04-03
changed the style of feats

bleehu:
so that the name of each of the feats shows up in the table of contents.
screenshot from 2018-04-03 16-44-47


Turtlelord26 (approved):
Everything seems to be in order, 👍 

I was right, this does make it easier to scan prereqs in github, let alone quick referencing via docs.

PR #421, opened 2018-04-04
Mh issue 130 lv10 ammo

bleehu:
Should add the ammo that would fix Issue/Feature Request #130. The hope is that these special ammos will offer fun loot for later dungeons, offer fun differences in kind and create the opportunity to use different ammos as a tactical puzzle (similar to the plasma/bullet dichotomy in Halo games).


Turtlelord26 (approved):
👍

PR #422, opened 2018-04-04
Changed the style of weapon attachments.

bleehu:
This should go from this....

old attachments

To this...

new attachments


Turtlelord26 (approved):
Oh my. Such parsability and style compliance.

Issue #376, opened 2018-03-15
Gripe: Phrasing of "Medicine Saving Throws"


rosebrallier:
https://github.com/bleehu/Compound_X/blob/c3463269fc0d0c8ee2c7f788ba95ccae8a298e3d/01_01_Basic_Rules.txt#L567

Using the title "Medicine Checks" or "Medicine Rolls" would better describe the content of the section, especially since Saving Throws are their own mechanic in the game (and "Medicine" is not one of them).

Turtlelord26:
I'm thinking that that whole section should be looked at and redone (especially that heal skill calculation) as a Hit Point Recovery section that includes rules for healing over time via rest and includes the Doctor class only by way of mentioning that they have alternative means of healing people.

It also might live more happily in a different location of the rules.

Related: The current "Resting" section is no longer in aesthetic, as it only talks about nanites.

PR #424, opened 2018-04-04
removed reference to Medicine Saving Throws

bleehu:
Should be an answer to what was agreed upon in Issue #376. I agree that there should be some additional design on healing, but until we have specific intentions there, I'd like to not start the work.


Turtlelord26 (approved):
A successful purge of the misuse of saving throw. Though by itself it does not address all points raised in the cited issue, it does fix the core gripe.

Issue #364, opened 2018-03-08
Melee 2.1 Part 2: Nonlethal Damage


Turtlelord26:
See #356 

The whole nonlethal damage section would benefit from a clarity pass. 
The following changes are requested:
- [x] Remove the nonlethal health bar
- [x] allow individual melee attacks to be declared nonlethal
- [x] implement an accuracy penalty to nonlethal attacks, as it is harder to do damage when you're trying to not do damage
- [x] explicitly state that characters downed by nonlethal attacks are unconscious and automatically stable.
- [x] remove damage penalty to unarmed attacks (damage of unarmed vs unarmed (trained) can be handled in the equipment section).

PR #449, opened 2018-04-12
Melee 2.1 Nonlethal Damage Update

Turtlelord26:
Solves Issue #364 

The exact accuracy penalty for nonlethal attacks is testable and deliberately has anti-synergy with "Hitting what you intend", since that mechanic flavors to hitting extra-vital areas and nonlethal attacks are the opposite.


bleehu (approved):
Well this is so much easier to understand and play that I don't know why it was so much more complicated in the first place. You've made design look easy and obvious. Very well done!

PR #450, opened 2018-04-12
Turtlelord26 fix dev commit trash

Turtlelord26:
Apologies for inflicting this review upon you. For ease of comparison, below is the commit that encompasses all of the unwanted (at least, at this time) changes to dev.

https://github.com/bleehu/Compound_X/commit/6ef5b9223cb0feea89d604bfd5884abab4e0b936

### DIFF
--- a/CharacterCreation/02_Classes.txt
+++ b/CharacterCreation/02_Classes.txt
@@ -736,7 +736,7 @@ Class Primary Attribute: Strength
 Starting Wealth: $2170 + 100*Luck
 
 Passives:
-* Gain Feat: Unarmed Combat Training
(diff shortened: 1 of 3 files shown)


bleehu:
You want to un-add the melee update? I forgot, did we review and greenlight the melee feats? If so, shouldn't we keep it in dev?

Turtlelord26:
That merge is still unapproved, but regardless it's going to the project superbranch. In order to keep dev as internally consistent as possible, I don't want to push a partial project to it. Hence the superbranch.

bleehu (approved):
We all good. shipit! Or un-ship it as the case may be. Truck it?

Issue #372, opened 2018-03-15
Gripe: Edit Skill Requirements and Class Passives to reflect Systematized Skills


rosebrallier:
For example:
Using grenades requires 10 skill in explosives, which is no feasible in early levels because skill points at lvl1 have decreased by 50.
Pilot passively gives 10 in piloting (a skill which no longer exists).

bleehu:
@rosebrallier can you handle this? Do you have enough familiarity with Git to make it so?

rosebrallier:
This is on my docket after I get reviews and edits on the systematized skills update.

rosebrallier:
Merged or PR'd:
- [Basic Rules](https://github.com/bleehu/Compound_X/pull/432)
- [Classes](https://github.com/bleehu/Compound_X/pull/410)
Still to come:
- Feats
- Weapons Docs

rosebrallier:
Note to self: remember to consistently identify specific skill checks (ie "DC of Athletics checks to break down a door" as opposed to "door-breaking DC")

rosebrallier:
There may be further instances to clean up, but the full repo has been reviewed for skills consistency as of PR #579.

PR #452, opened 2018-04-12
skills consistency feats

rosebrallier:
Changed defunct skills to their new equivalent
Changed formatting conventions for skills
Edited some feats' skills-relevant descriptions for clarity or style
Pursuant to Issue #372

### DIFF
--- a/CharacterCreation/03_Feats.txt
+++ b/CharacterCreation/03_Feats.txt
@@ -307,10 +307,10 @@ damages an area, Last Shot Counts does not take effect.
 Prerequisites:
 *  Pilot or 40 skill in Vehicles - Flight OR Space
 
-	You gain a +20% to Vehicles checks in either airborne or spaceborne vehicles
-(choose one), and an additional +20% to unconventional maneuvers in such as well,
-but you and your vehicle take double damage for the next three turns after using these
-bonuses. (any teammates in such a vehicle do not take double damage)
+	You gain a +20% to Vehicles checks in airborne or spaceborne vehicles, and an
+additional +20% to unconventional maneuvers in such as well, but you and your vehicle
+take double damage for the next three turns after using these bonuses. (any teammates
+in such a vehicle do not take double damage)
 
 == Lucky Blind Fire ==
 
--- a/CharacterCreation/03_Feats.txt
+++ b/CharacterCreation/03_Feats.txt
@@ -242,7 +242,7 @@ action by 30%, (30% better roll, +30% damage [drop the remainder], up to you.)
 
 Prerequisites:
 
-*  Pilot or 20 skill in Piloting
+*  Pilot or 20 skill in Vehicles - Ground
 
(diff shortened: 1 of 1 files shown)


CharacterCreation/03_Feats.txt
```diff
@@ -305,11 +305,11 @@ damages an area, Last Shot Counts does not take effect.
...
 
 Prerequisites:
-*  Pilot or 40 skill in piloting
+*  Pilot or 40 skill in Vehicles - Flight OR Space
 
-	You get a +20% to pilot skill checks in airborne or spaceborne vehicles, and
-an additional +20% to unconventional maneuvers in such as well, but you and
-your vehicle take double damage for the next three turns after using these
+	You gain a +20% to Vehicles checks in either airborne or spaceborne vehicles
+(choose one), and an additional +20% to unconventional maneuvers in such as well,
```
Turtlelord26: I wouldn't mind giving both with the one instance of the feat (would also consider explicitly stating that you can take the feat twice to be able to apply it to both vehicle types). I don't think I like getting one and forever locking out the other.

CharacterCreation/03_Feats.txt
```diff
@@ -124,8 +124,8 @@ Prerequisites:
 *  None
 
 	Make a stealth check to hide your chosen pistol, shotgun, or SMG on your 
-person. To succeed, you need a minimum score of 20. Roll determines DC for 
-enemy to actively search and find your hidden weapon.
+person. To succeed, you must pass a DC 19 check. Total roll determines DC for an
```
Turtlelord26, line 114: Aren't skill checks success on tie? DC not-divisible-by-5 is very unusual.

Turtlelord26 (changes requested):
👍

CharacterCreation/03_Feats.txt
rosebrallier: To quote basic rules: "Anytime you 
need to Pass a Skill check, there will be a DC (Difficulty Check) you have to 
exceed in order to successfully pass the check."
I'll happily change it back to 20.

Turtlelord26: Nope, you're correct. DC 19 is the equivalent and more style-consistent phrasing. Disregard.

bleehu (approved):
I too would like to see some small changes to leaf on the wind that Turtle suggested, but I see those changes as non-blocking. Thanks a bunch for cleaning up these details! You're really helping to sanitize the confusion out of our rules! Shipit!

rosebrallier:
I swapped Leaf back to giving both space and air. After thinking about it a little, if you /want/ to use Leaf in a vehicle you don't have a ton of skill in, I figure the risk balances out the benefit.

Turtlelord26 (approved):

Turtlelord26:
There is also the possible confusion and rules lookups from the particular case of flying a spaceship through atmosphere that that change nullifies 👍

PR #460, opened 2018-04-14
Turtlelord26 #398 weapon expertise feats

Turtlelord26:
Solves Issue #398 

"Proficiency skills" are nuked, weapon expertise feats now reference existing weapon skill. They no longer give skill points, but to compensate the first benefit comes at 0 skill points.

Also deleted the stale skills doc and put its name on the new skills doc.

**Note: I did not rebalance the benefits from these feats. That will have to be done at some point.**


rosebrallier (approved):
YAY

bleehu (changes requested):
Hang on, you totally changed the balance on the feats. Not only did you remove mastery of each, but you also more than doubled the amount of skill points needed to get to each next buff. This is taking something playable and making it unplayable. Furthermore, removing the skills gained from the feats further renders them less useful, as you can no longer use the feats to get better at reloading your weapon under fire. I recommend aborting this change entirely and making a separate PR for unifying the systematized rules.

Turtlelord26:
Update, balance is now happening on this branch. Will close this PR and reopen it after some retouching to the feat bonuses aimed at keeping them playable.

PR #463, opened 2018-04-15
Mh/partial play update

bleehu:
How's that? I got it down to 197 commits. This doesn't quite catch us up to December. I should be able to pick arbitrary points of the timeline to jump to now using the 
`$git reset --hard <timestamp> `
command. 

Another valid strategy might be to just shove all of R_C to play and then have @Wildlovelies proofread it?


Turtlelord26 (approved):
Big chunks of stuff this puts in play includes
-Secondary stats
-The death/dying/downed section
-Literally Combat 2.0
-Modern class doc formatting
-Removed Mage and Sorc
-Feat cleaning and c2.0 compliance, some actives moved from passive to active list
-introduced attribute-based feats
-weapons moved to individual documents separated by type.
-combat 2.0 armor
-skills reorg
-char sheet files deleted
-medic procedure spellcheck and graphic update
-Paladin spell cost rebalance
-added Glossary

No idea if paladin spells were actually tested but given their ultimate fate I don't think it matters all that much.

c2.0 armor has unfortunately not been tested (correctly), but c2.0 is unplayable without it so I still say ship that. It's been working out well enough despite our error that I have faith enough to put it in play while we iterate further towards static armor class.

Issue #400, opened 2018-03-25
Feature Request: Skill(General) buff


Turtlelord26:
From #385 

When dealing with progression systems, I always try to aim for maximum end-state power. This involves lots of turtling in RTS and 4X, for analogy. The following concern assumes that mindset is common enough to design around.

I am wary of ever putting points into General skills, because at some point General must become overshadowed by the Specific skills (and if it doesn't we've made a design error because then the specifics have no reason for existing). I wonder if we could mitigate that by having a threshold where having general skill helps your specific skill rolls. Something like having 30/60 points in the General lets you roll the average on a Specific check that you are not using your general score for (including in combat situations), and 60/60 rolls with advantage.

rosebrallier:
This is a cool idea to take into account how in real life, knowing about a General subject makes it easier to approach subsets of that subject (since you have the contextual knowledge and general thinking framework for the field).
What do you think @bleehu ? Useful buff of complexity creep?

rosebrallier:
ping

bleehu:
I'm concerned that's going to make general skills too valuable.  The "wildcard bonus" of being able to use general skills in multiple skill challenges is already a huge equalizer. I'd like to discuss this more,  but I hesitate to write a wall of text. Take it up in person this weekend?

Turtlelord26:
Consensus has been reached that General skills are testable in their current iteration, with the exceptino of Knowledge(General), which shall be nerfed to specify that when a character takes Knowledge(General), they must declare 4 areas of specific knowledge that the General check may apply to, and cannot apply the general check to other specific knowledges.

PR #469, opened 2018-04-16
Knowledge - General nerf

rosebrallier:
Narrowed the wheelhouse of Knowledge - General
Re Issue #400

### DIFF
--- a/CharacterCreation/06_Skills.txt
+++ b/CharacterCreation/06_Skills.txt
@@ -44,7 +44,8 @@ Vehicles - General to the roll.
 	Knowledge - General is a minor exception to this rule because Knowledge
 is an incredibly broad field. When a character learns the Knowledge - General skill,
 the player must choose four specific topics that the character has some experience
-or education in. Knowledge - General can apply to rolls related to those four topics.
+or education in. Knowledge - General can apply only to rolls related to those four
+topics, at the discretion of the GM.
 
 ===== Skills in Compound X =====
 	This list of some of the most commonly-used skills in Compound X is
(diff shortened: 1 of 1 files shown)


bleehu (approved):
shipit!

CharacterCreation/06_Skills.txt
```diff
@@ -36,11 +36,15 @@ For an extra critical, add 20.
...
-Knowledge - General to the roll.
+group (rounded down). For example, if a GM asks for a roll in Vehicles - Flight,
+but the character only has 10 skill in Vehicles - General, the player can add 5
+from Vehicles - General to the Vehicles - Flight roll. However, if the  GM asks
+for a roll in Vehicles - General, the player can add all 10 points from
+Vehicles - General to the roll.
+	Knowledge - General is a minor exception to this rule because Knowledge
+is an incredibly broad field. When a character learns the Knowledge - General skill,
+the player must choose four specific topics that the character has some experience
+or education in. Knowledge - General can apply to rolls related to those four topics.
```
Turtlelord26: Not a blocker but for this to technically mean what we want it to mean we need to have "can apply ^only^ to rolls related..."

Turtlelord26 (commented):
Should we make it clear that the "topics" are (or are not?) the enumerated specific knowledge skill fields?

CharacterCreation/06_Skills.txt
rosebrallier: I agree on adding an "only," but I don't want to say that they must choose from the list of specific knowledge skills. This is because my skills list is not meant to be exhaustive.

Turtlelord26 (approved):
That has sense 👍

Issue #464, opened 2018-04-15
Reload check crits


Turtlelord26:
There's a line in Basic Rules that states that skill bonus is taken into account for determining critical, instead of determining criticals solely on the raw roll. Isn't that wrong?
https://github.com/bleehu/Compound_X/blob/dev/01_01_Basic_Rules.txt#L1099

1sourcecontrol:
Took me a bit to find the commit, but I believe I made this change 9 months ago. The skill-adjusted critical roll was the intended effect, to fulfill 2 goals:

1. Strengthen weapon use (x) skill. I wanted weapon use skill to fulfill two roles. Reducing the probability of 'bad reloads' and increase the probability of critical good reloads. --Note, I don't remember realizing this back then, but having weapon use(x) skill affect critical reload probability was one of the few reasons to take weapon use(x) skill beyond the minimum. I.e., if your favorite weapon had a reload DC of 30% there weren't many reasons (or any reasons?) to take weapon use(x) skill beyond 30. 

2. The Feat "First Shot Counts" was dependent on critical reloading, so the skill added to critical would help improve the usefulness of the feat(s). Although, this is no longer the case because the feat has been reworked. 
-- But the Feat rework basically includes the same concept

If I recall correctly.

Turtlelord26:
Those are valid goals but I would prefer we fulfill them without creating hard-to-remember exceptions to our rules. Weapons - [X] skill power is definitely on that radar.

1sourcecontrol:
By all means :+1:

bleehu:
Am I still needed on this? (trying to keep organized on what's assigned to me and what isn't)

Turtlelord26:
Only if you would like to review the PR

1sourcecontrol:
So, granting the merge of PR 475, does this mean that CX is back to: "I won't be getting a weapon expertise feat, so there is _no_ reason to get weapon use(x) beyond my highest weapon's reload DC"?

Turtlelord26:
Adding a note on #453 to add a DC chart for Weapons - [X] skills would help us codify all of the things those skills are for.

rosebrallier:
Closing this issue as resolved; questions of the balance of Weapon - [x] skills belong in their own issue and may be at least partially resolved via Issue #398

PR #475, opened 2018-04-20
Remove skill from crit range in reloading

Turtlelord26:
#464
Now a regular Weapons - [X] check, which crits like any other skill check.


bleehu (approved):

rosebrallier (approved):
Jexcellent!

PR #479, opened 2018-04-23
bumped version number

bleehu:


Turtlelord26:
Why does the version bump PR have 59 commits?

bleehu:
Among other things, it deletes the human doc, which is a change that RC got that dev should not. My hunch is that this change has RC as a parent when I had intended it to have dev as the parent.

Turtlelord26:
Does that mean this needs some doctoring before it gets reviewed?

bleehu:
Probably doctored like you doctor a horse with a broken leg...

Reveraine (commented):
I'm a little confused about this. 
While it's true the races are not ready for play, are you not merging them to dev?

Turtlelord26:
Species are already in dev and not intended for deletion there. That's part of the reason this PR is messed up.

bleehu:
We're not-not merging them to dev; this is a bad PR that is aborted because it's wrong heritage created that delta which would remove the Species work from dev.

PR #481, opened 2018-04-24
added tricks

bleehu:
Obviously this is a _very_ abreviated list that exists mostly to show the shape of what I'm thinking of doing for the progression of the dog. Thoughts? Questions? Comments? Snark?

I need to backfill some documentation on the design repo to show where these are coming from and why I'm doing things this way. But before I do that, I want to see what the design reaction is to the as-built idea.

### DIFF
--- a/CharacterCreation/Class-Specific Documentation/Hunter Dog Tricks.txt
+++ b/CharacterCreation/Class-Specific Documentation/Hunter Dog Tricks.txt
@@ -1,0 +1,63 @@
+==============================
+Hunter's Dog Tricks
+==============================
+
+    When a Hunter levels up, so does their pet. Every time a pet levels, they
+unlock a Trait. Most Traits come with a bonus to Courage, Toughness, or 
+Playfulness. Courage, Toughness, and Playfulness are cumulative. Whenever a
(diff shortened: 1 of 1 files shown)


CharacterCreation/Class-Specific Documentation/Hunter Dog Tricks.txt
```diff
@@ -0,0 +1,63 @@
...
+
+== Goofball ==
+Prerequisites:
+*  Level 5 or higher
+
+Bonuses:
+*  Playfulness + 15
+
+    Your pet is a goofball! They love to have fun and play around. They learn
+tricks that have to do with
```
Reveraine, line 40: Unfinished sentence?

Reveraine (approved):
Perhaps unfinished sentence. Potentially non-blocking if this continues to be revised.

CharacterCreation/Class-Specific Documentation/Hunter Dog Tricks.txt
bleehu: Will fix on the bus.

PR #482, opened 2018-04-24
added some rules about attack dogs

bleehu:
Here is the very beginning of the rules for having a dog. Stats for the dog will be added in a separate small change, just in case there are differing concerns over stats and rules. 

Changed CharacterCreation/02_Classes.txt

### DIFF
--- a/CharacterCreation/02_Classes.txt
+++ b/CharacterCreation/02_Classes.txt
@@ -392,6 +392,7 @@ Passives:
 * Purchase your first primary cloaking module at a $1500 discount
 * Cloaking's Nanite Activation cost is reduced by 4 after level 3
 * Gain Ability "Mark Quarry" Detailed below
+* At Level 3, gain an Attack Dog, detailed below
 
 Gain Ability "Mark Quarry": Lvl/2+1 Uses per day - As a free action, expend 20 nanites to Mark 
 a visible target. You must be able to visibly see the target or something the target is wearing. 
@@ -415,6 +416,24 @@ Skill Proficiencies:
 * Weapons - [Choose one]
 * Spotting
 * Survival
+
+== Hunter's Dog ==
+
+    When your character ascends to level 3, you receive an attack dog. Maybe 
+you found the dog along your last mission, or maybe a dog was issued to you when
+you returned to home base (at the GM's choice). Every time you level up after
+you have your dog, your dog levels up as well and learns a new "Trick" as 
+defined in the "Class-Specific Documentation/Dog_Tricks.txt" document. For 
+simplicity, Dogs start at level 3 so that they are always the same level as 
+their owners. 
+
+    Your new dog is well-trained and responds to a variety of commands. Your dog 
+listens to you and does his best to make his or her owner happy, but is not 
+controlled by the player. Instead, the owner character commands the dog, and the
+GM controls how the dog responds. 
+
+Dog Personalities:
+
 
 
 == Class Feats ==


Reveraine (approved):
Ship It!

CharacterCreation/02_Classes.txt
```diff
@@ -416,6 +417,24 @@ Skill Proficiencies:
 * Spotting
 * Survival
 
+== Hunter's Dog ==
+
+    When your character ascends to level 3, you receive an attack dog. Maybe
```
rosebrallier, line 422: Wording note: I like "reaches" or "levels up" much more than "ascends." Different connotations.

CharacterCreation/02_Classes.txt
```diff
@@ -416,6 +417,24 @@ Skill Proficiencies:
...
+you returned to home base (at the GM's choice). Every time you level up after
+you have your dog, your dog levels up as well and learns a new "Trick" as 
+defined in the "Class-Specific Documentation/Dog_Tricks.txt" document. For 
+simplicity, Dogs start at level 3 so that they are always the same level as 
+their owners. 
+
+    Your new dog is well-trained and responds to a variety of commands. Your dog 
+listens to you and does his best to make his or her owner happy, but is not 
+controlled by the player. Instead, the owner character commands the dog, and the
+GM controls how the dog responds.
```
rosebrallier, line 433: I worry a little about increasing the burden on the GM by giving them more in-game work.

rosebrallier (commented):
I like it overall; left a few detail notes/questions.

CharacterCreation/02_Classes.txt
bleehu: ahah, I do very much too. Further more I worry very much about a new player projecting a beloved pet onto the dog and absolutely losing it if the dog dies in the line of duty. I've seen something similar go horribly wrong in another RPG back in high school. 

I'm so concerned with these two issues that I'm considering halting work on Hunter and leaving it how it is. Or maybe doing only minor touch-ups and scrapping the dog idea altogether.

Turtlelord26: The work flow problem at least we might ameliorate by giving the player nominal control over the pet but with explicit DM veto power.

As for the other bit, would it be enough to include a projection warning in our DMG-equivalent 'handling players' section?

I believe we have solvable design problems here. Pets are good.

Turtlelord26: Oh, another thought. Maybe give the dog bleeding out rules like PCs so it can be saved if it goes down.

bleehu: My intention is to give specific language saying that the dog is well-trained and friendly to both the handler and their teammates, to prevent knuckle-heading "My dog is antagonizing you, but it's not me or my character" nonsense. I think we'll need warnings in both the DM guide and Player handbook on all of these issues, including the potential danger for the dreaded DMPC-diva.

bleehu: I agree that these design problems are solvable. but I would liken it to flying a little close to the sun. It's interesting and unique, but volatile.

bleehu:
I think I'll call that a tentative "shipit" to send it to the superbranch and let Turtlelord get his required approval or veto during the superbranch->dev merge.

CharacterCreation/02_Classes.txt
Turtlelord26: @rosebrallier How did Critical Role handle the friendly Trinket?

rosebrallier: I didn't get to answer this question b4 merge :(
Trinket was controlled by Vex/Laura, but please don't use 5e ranger w/ pet as a template.

Turtlelord26: What's wrong with 5e ranger w/ pet? (I honestly don't know). Avoiding whatever they did wrong sounds like a good design choice.

rosebrallier: In 5e, the pet scales terribly. At higher levels (like 10+), it's attack rolls, hp and dmg become inconsequential and unimpactful to the point where it's attacks (which require the player's action) simply can't compete even with the ranger's underpowered spells. The pet can generate sneak attack (flanking) bonuses and not much else. Also, the pet has no out-of-combat function.

bleehu: At the risk of hubris, I think I can out-design that handily. I will not be using 5E (nor 4E) as a template.

Turtlelord26: Excellent

PR #490, opened 2018-04-29
Turtlelord26 Melee 2.1 Attack Calculations

Turtlelord26:
As usual, design info in my commit notes.

    I think randomness is still a heavier factor than it was when skill was directly added.
    -Basically I decided to add both STR and DEX combat mods to the accuracy roll, 
    since you can just smash your way through an attempted block and the most 
    effective brawlers will use both techniques. I almost did the same thing to Guard 
    but decided to have FORT oppose STR. This is very open to discussion.
    -What isn't seen here is that I plan to have the Melee Weapon and Unarmed 
    Expertise feats also contribute to Accuracy and Guard rolls.
    -Finally, the weapons themselves will contribute. I've decided to let weapons 
    keep their individual accuracy and have Guard piggyback on that (with certain 
    exceptions, like whips, that should contribute far less to defense than offense). 
    Other weapons got Guard modifiers as shown in the friendly table at category 
    level, to keep things relatively simple.

    We almost certainly will need a balance pass on these numbers and on 
    individual melee weapon accuracy mods if this goes through.

What do people think of this implementation?


Items/04_01_Melee_&_Thrown_Weapons.txt
```diff
@@ -1,417 +1,418 @@
...
+==============================
+
+	These weapons/attacks use an 'Accuracy Modifier' instead of
+a 'Miss Chance', which is added to your combat dex modifier and
+and any relevant feat or skill bonuses to oppose your opponent's 
+Guard Class, or GC. Melee weapons do not pierce armor, unless otherwise 
+stated.
+
+	Melee weapons come in one of four size categories. The first 
+three are Light, One-handed, Two-handed. Characters start
```
rosebrallier: nitpick: "and two-handed"

Items/04_01_Melee_&_Thrown_Weapons.txt
```diff
@@ -1,417 +1,418 @@
...
+Guard Class, or GC. Melee weapons do not pierce armor, unless otherwise 
+stated.
+
+	Melee weapons come in one of four size categories. The first 
+three are Light, One-handed, Two-handed. Characters start 
+with proficiency with some number of these based on their class. The 
+fifth category is Exotic. Acquiring proficiency with Exotic weapons does 
+not grant access to all Exotic weapons, but rather to a particular type, 
+such as whips. Exotic proficiencies may rarely be granted by Backgrounds, 
+and are more commonly acquired by feat.
```
rosebrallier: "feats" or "a feat"

Items/04_01_Melee_&_Thrown_Weapons.txt
```diff
@@ -1,417 +1,418 @@
...
+defensive implement in his or her off-hand, that item cannot be used and 
+it provides no bonuses for one round.
+
+==============================
+UNARMED
+==============================
+
+== Unarmed Strike ==
+
+Damage: 20 + STR mod + DEX mod, minimum 10
```
rosebrallier, line 53: Can you explain the minimum 10 to me?

Items/04_01_Melee_&_Thrown_Weapons.txt
```diff
@@ -1,417 +1,418 @@
...
+
+Damage: 50 + 6*STR
+Size: Two-handed
+Range: 2m
+Accuracy Modifier: +0
+STR Requirement: 6
+
+	Considered concussive (explosive) damage. Hits armor points 
+before physical damage if the target's armor has defense against explosive (i.e. 
+concussive) damage. If armor roll is greater than 50%, target rolls a DC 50
```
rosebrallier: We should choose whether to standardize to either concussive or explosive imo

Items/04_01_Melee_&_Thrown_Weapons.txt
```diff
@@ -1,417 +1,418 @@
...
+(non-Nanite) shield, all plasma burn is canceled, the shield takes 30 damage and 
+the plasma sword passes through the shield.
+
+
+==============================
+THROWN WEAPONS
+==============================
+
+	These weapons work similarly to guns. However your range is determined by 
+your DEX or STR and your miss chance is reduced by your [Combat PER Mod]. Like
```
rosebrallier: Sentence structure: since miss chance with guns is also reduced by combat Per mod, "your miss chance is reduced by your [Combat PER Mod]" belongs in the previous sentence (or any sentence except for one qualified by "however")

Items/04_01_Melee_&_Thrown_Weapons.txt
```diff
@@ -1,417 +1,418 @@
...
+the plasma sword passes through the shield.
+
+
+==============================
+THROWN WEAPONS
+==============================
+
+	These weapons work similarly to guns. However your range is determined by 
+your DEX or STR and your miss chance is reduced by your [Combat PER Mod]. Like 
+melee attacks, if you roll a Crit or 3+ above your Miss Chance, you ignore the
```
rosebrallier: Is it a consistent convention to abbreviate critical success with "crit"? I don't actually know if we've decided on this.

rosebrallier (approved):
A few grammar notes and some questions.

rosebrallier:
Also! I really appreciate the Hybrid example (idr if I've told you that or not yet.)

Items/04_01_Melee_&_Thrown_Weapons.txt
Turtlelord26: An unarmed strike can never have outgoing damage less than 10. It's to prevent damage underflow from negative attribute mods.

Another option is to write a blanket rule that negative mods don't subtract damage from any melee weapon, but I rather like the idea that a 2 STR character is worse at swordplay than a 5 STR character.

Items/04_01_Melee_&_Thrown_Weapons.txt
Turtlelord26: I agree completely, and that's not the only problem with this weapon (see outdated reference to door-kicking skill). Unfortunately, the repository reorg interacted oddly with the commits on this branch, which I forked before the reorg, so the entire file is green instead of just the changed bits.

That's a long way of saying that that's outside the scope of this project.

Turtlelord26: Want to make an Issue?

Items/04_01_Melee_&_Thrown_Weapons.txt
Turtlelord26: Another artifact from the original document, but I'll get this one while doing the other typography.

To actually answer: no, we don't have the "Crit" abbreviation codified in writing.

Items/04_01_Melee_&_Thrown_Weapons.txt
Turtlelord26: Also an artifact from the original document, but since it's typography I'll terminate it here.

Items/04_01_Melee_&_Thrown_Weapons.txt
rosebrallier: Oh, I understand having a minimum, I'm just wondering why it's 10, and not 5 or 0 or w/e. Is it so anyone can throw a punch in a pinch, instead of being in a situation of "welp it's not even worth an action to try to punch"? I think I answered my own question...

Items/04_01_Melee_&_Thrown_Weapons.txt
rosebrallier: 👍  On it.

Items/04_01_Melee_&_Thrown_Weapons.txt
Turtlelord26: Mea culpa, wasn't playing attention when I deleted the fourth item in the list.

Items/04_01_Melee_&_Thrown_Weapons.txt
rosebrallier: What reference to door-kicking? I'm only seeing reference to door-breaking DC which is a legit subset of STR checks.

Items/04_01_Melee_&_Thrown_Weapons.txt
Turtlelord26: Honestly, besides putting the minimum above zero, very little conscious thought went into picking ten. I guess it's just a friend-shaped round number in a d10-based system.
But yes, nonzero minimum for basically that reason.

Items/04_01_Melee_&_Thrown_Weapons.txt
Turtlelord26: Fair enough. I would prefer we templated skill check references like 
"[Skill/Attribute Name] check to [Perform Action]"
instead of "[Perform Action] DC", which in this case would look like "Athletics-STR (or Strength) DC to break down a door"

In this case it looked like a reference to the defunct door-kicking skill that I recall we folded into Athletics.

rosebrallier: I'll make a note to take care of this in Issue #372

Items/04_01_Melee_&_Thrown_Weapons.txt
rosebrallier: I'll make an Issue!

rosebrallier:
In re-reading the PR, I realised that I didn't actually explicitly answer your original intent. I like the implementation.

bleehu (changes requested):
I hate to do this, but could you add your changes in a new branch that has the information in the new combat rules file?

Turtlelord26:
That wasn't so bad.

Although, you can't tell what was changed from the aggregate files changed tab at all anymore. I recommend looking at individual commits.

rosebrallier:
@bleehu ping!

bleehu (approved):
shipit!

PR #515, opened 2018-05-22
added new feat: kip up

bleehu:
Seemed like a fairly obvious missing feat. Needed a break from craziness at work and with wedding planning. So I got a little creative.

Not sure about the prereqs, I'm not positive we're giving more useful choices for characters that have particular stats. 

The nanite cost seemed like a good rate-limiter to keep someone from abusing this to do the worm.

### DIFF
--- a/CharacterCreation/03_Feats.txt
+++ b/CharacterCreation/03_Feats.txt
@@ -863,6 +863,25 @@ additional attacks, where X is your Nanites spent/5. Each additional attack has 
 compounding +2 miss chance. (A 10th level character making their 3rd strike in a 
 row would be rolling at +4 miss chance).
 
+== Kip-up [15 Nanites] ==
+Prerequisites:
+
+* DEX 6
+* STR 5
+
+	A Kip-up is a gymnastic maneuver in which a character lying on the ground
+curls up and then kicks up and out into the air, using the resulting momentum to 
+jump from their back straight to their feet. This kick does not work well as a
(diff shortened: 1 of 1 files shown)


Turtlelord26 (commented):
What would you think of instead having the prereq be points in Acrobatics?

dfuchs2:
Good idea. My only question is about nanite usage. Aside from keeping people from worming, is there a reason for the ability to use them? Maybe there could be a cooldown of 3 turns in which the player receives a -20 to acrobatics or something.

Turtlelord26:
Such a cooldown would be enough without adding any skill penalty methinks.
@dfuchs2 What if it has no baseline nanite cost, a cooldown, and a nanite payment to ignore the cooldown 1/turn.

bleehu:
My concern with adding a cooldown longer than just 1 per round is that I'm a little concerned about giving the player too many different things to keep track of. Other than that, you're right. There isn't much reason to use nanites.

dfuchs2:
I like @Turtlelord26's idea. The ability to choose to ignore a cooldown is an interesting idea.
Also, I agree that 3 turns is too long. Too much to keep track of. I'd be fine with it only being 1 turn. It is a feat after all. Players don't get very many of them and they shouldn't be restricted heavily.

bleehu:
How is that? Please review! (like, hit the green review button!)

dfuchs2 (approved):
Looks good to me. I could definitely see myself using it if I build a high dex character. Maybe an assassin character or something.

PR #522, opened 2018-06-12
Spelled out colloquialism "crit"

bleehu:
into critical hit, etc, in classes and weapons docs

 Changes to be committed:
	modified:   CharacterCreation/02_Classes.txt
	modified:   Items/04_00_WeaponAttachments.txt

### DIFF
--- a/Items/04_00_WeaponAttachments.txt
+++ b/Items/04_00_WeaponAttachments.txt
@@ -606,9 +606,9 @@ round where it's gonna hurt.
 
     Vorpal rounds deal an extra 10% of normal weapon damage on critical hit and
 20% damage on extra-crits. For example, if your gun deals 80 damage normally,
-and you roll a crit when using Vorpal rounds, your gun deals 80 + 10 (for crit)
-+ 8 (Vorpal damage) = 98. If you scored an extra-crit with the same weapon, 
-you'd do 80 + 20 (extra crit damage) + 16 = 116 damage. 
+and you roll a critical hit when using Vorpal rounds, your gun deals 80 + 10 
+(for crit) + 8 (Vorpal damage) = 98. If you scored an extra-crit with the same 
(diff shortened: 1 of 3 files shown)


Items/04_00_WeaponAttachments.txt
```diff
@@ -606,9 +606,9 @@ round where it's gonna hurt.
 
     Vorpal rounds deal an extra 10% of normal weapon damage on critical hit and
 20% damage on extra-crits. For example, if your gun deals 80 damage normally,
-and you roll a crit when using Vorpal rounds, your gun deals 80 + 10 (for crit)
-+ 8 (Vorpal damage) = 98. If you scored an extra-crit with the same weapon, 
-you'd do 80 + 20 (extra crit damage) + 16 = 116 damage. 
+and you roll a critical hit when using Vorpal rounds, your gun deals 80 + 10 
+(for crit) + 8 (Vorpal damage) = 98. If you scored an extra-crit with the same
```
Turtlelord26, line 610: I think we also want extra-crit to be spelled out to extra-critical?

Turtlelord26 (changes requested):
Looks mostly excellent. 👍

Turtlelord26:
Found the changes that were blocking automatic merge, thought I'd take care of that since I made them.

Items/04_00_WeaponAttachments.txt
bleehu: So we actually don't define "extra-critical" but instead formally define "extra-crit" in the combat rules...

Turtlelord26: Oh. That's a consistency problem. We should change that. But it's also different enough that we can classify it as a new issue.

Turtlelord26 (approved):

Issue #505, opened 2018-05-14
Gripe Query: Flash-bang Reflexes


Turtlelord26:
Flashbangs have PER saves to throw off the effect, but no provision for Reflex saves negating or reducing it entirely up-front. Imagine seeing the grenade incoming and reacting fast enough to cover one's eyes.

PR #523, opened 2018-06-12
Added Reflex Save to grenades

bleehu:
moved Reflex Modifier to stat block of grenades and added a reflex save to all grenades
Issue #505 


Turtlelord26 (changes requested):
We need to specify what making (and failing) the reflex save means for each grenade.

Turtlelord26 (changes requested):
At ground zero of the 10m cloud grenades, making the reflex save lets you dive 5m. That's a long way. Not sure if flavor compatible. 
What if instead the reward for making the save is diving prone behind cover (relative to ground zero), *if such cover is available nearby*, and otherwise hunkering in place to take somewhat reduced damage?
At the risk of complexity, we could also allow diving onto the grenade to provide everyone around you cover at the cost of massively increased ouch.

bleehu:
I think it's fairly intuitive to add some complexity at this point. I'll do that.

Diving on a grenade sounds like maybe a feat? Call it "courage" or something? What do you think? That way we spread the complexity around a bit

Turtlelord26:
I'm not confident that a feat for diving onto grenades would ever be taken over other feats less likely to kill people's characters. If it was a feat I recommend it be stapled to another effect, and secondary to it.

dfuchs2 (approved):
I like it. It's good to remove the potential for movement exploits.

Turtlelord26 (approved):

PR #525, opened 2018-06-13
added "return to sender" feat

bleehu:
A new feat that answers the question, why can't I do this normally?


Turtlelord26 (approved):
Sounds like we need a cooking mechanic with friend-shaped friendly friend-like catastrophic failure consequences.

Balance question: should we rate-limit this to once per round?

bleehu:
I figured the luck/REF saves were enough of a limiter.

PR #526, opened 2018-06-13
added overcharged ammo type for lasers

bleehu:
A very small amount of amelioration for the lack of fun attachments for lasers.

### DIFF
--- a/Items/04_00_WeaponAttachments.txt
+++ b/Items/04_00_WeaponAttachments.txt
@@ -521,12 +521,12 @@ that exceed the recommended voltage for compatibility with that gun. The result
 is a weapon system that imparts more damage through a higher intensity laser, at
 the cost of running the risk of overheating the emitter manifold and causing the
 weapon to "jam." Or in the worst case scenario, to overload even those safety
-measures and to cause the weapon slag, rendering the weapon inoperable and 
-potentially physically burning the user. This attachment is consumed upon use,
-and is only compatible with laser weapons.
+measures and to cause the weapon to slag, rendering the weapon inoperable and 
+potentially physically burning the user. This attachment is only usable with
+lasers.
 
 Cost: $1.5 x normal mag
-Damage: +15
+Damage: +15%
 Jam Range: + 2 on second roll
 
 == Poison Bullet Magazine  $2x + 50 magazine cost ==
--- a/Items/04_00_WeaponAttachments.txt
+++ b/Items/04_00_WeaponAttachments.txt
@@ -514,6 +514,21 @@ are excluded.
 == ED Electric Discharge Magazine  $2x + 20 magazine cost ==
     1/2 the bullet's base damage is converted to electrical damage.
 
+== Overcharge Laser Capacitors $(1.5 x mag cost) ==
+
+    Most common laser weapons have after-market ammunition batteries available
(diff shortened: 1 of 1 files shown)


Items/04_00_WeaponAttachments.txt
```diff
@@ -514,6 +514,21 @@ are excluded.
...
     1/2 the bullet's base damage is converted to electrical damage.
 
+== Overcharge Laser Capacitors $(1.5 x mag cost) ==
+
+    Most common laser weapons have after-market ammunition batteries available
+that exceed the recommended voltage for compatibility with that gun. The result
+is a weapon system that imparts more damage through a higher intensity laser, at
+the cost of running the risk of overheating the emitter manifold and causing the
+weapon to "jam." Or in the worst case scenario, to overload even those safety
+measures and to cause the weapon slag, rendering the weapon inoperable and
```
Turtlelord26: Typo nonblocker: weapon ^to^ slag.

Items/04_00_WeaponAttachments.txt
```diff
@@ -514,6 +514,21 @@ are excluded.
...
 
+== Overcharge Laser Capacitors $(1.5 x mag cost) ==
+
+    Most common laser weapons have after-market ammunition batteries available
+that exceed the recommended voltage for compatibility with that gun. The result
+is a weapon system that imparts more damage through a higher intensity laser, at
+the cost of running the risk of overheating the emitter manifold and causing the
+weapon to "jam." Or in the worst case scenario, to overload even those safety
+measures and to cause the weapon slag, rendering the weapon inoperable and 
+potentially physically burning the user. This attachment is consumed upon use,
```
Turtlelord26: aren't all magazine attachments consumed on use? possible redundancy flag.

Items/04_00_WeaponAttachments.txt
```diff
@@ -514,6 +514,21 @@ are excluded.
...
+and is only compatible with laser weapons.
+
+Cost: $1.5 x normal mag
+Damage: +15
```
Turtlelord26: Flat damage disproportionate affects miniguns (extreme case) and other fast-firing low-damage weapons. Intentional, or should we make it a % increase?

Turtlelord26 (approved):
Good, solid, functional, like. I'll see if I can get the tech team to subconscious a bit on what more we can do with lasers and plasmas.

bleehu:
ah, I totally agree with those. Made some changes; sorry to do this, but can I get a re-review? Specifically asking if a 15% boost to damage is balanced with the mag cost and jam range?

Turtlelord26 (approved):
I've dropped a numerical analysis into the slack. As expected, +jam chance wrecks autofiring guns like smgs. There's still an overall *dpa* increase, but factoring in clearing the jam there might not even be a *dps* increase. If there is one, it sure isn't cost effective. If we are ok with an ammo type not being very good with some types of fire, and potentially burning some players who don't run these kinds of technical analyses before buying, then this is approved. 
From a science standpoint, I would happily accept that the overcharged ammo plays better with slower fire.

There is some notable synergy between +jam chance as an ammo drawback and -jam chance from assault rifle expertise. Soldiers are going to soldier.

PR #527, opened 2018-06-13
Security Drone first pass

bleehu:
added semi-reasonable technobable for security drone and renamed file to have .txt extension

 Changes to be committed:
	new file:   01_02_Species_ECSC_Security_Drone.txt
	deleted:    01_02_Species_ECSC_Secutiry_Drone


Turtlelord26 (approved):
I'd buy it. I would consider additionally specifying that audio sensors are located on the head as well, but that is a nonblocking note.

PR #528, opened 2018-06-13
Mh/races push/refract

bleehu:
Added changes from first pass, and added a new refractory dermal layer for security drones as a new laser-resistance trait.


Turtlelord26 (approved):
[Stamps science approval]
I have to note again the flat damage disproportionately affecting some weapons, but in this case that may be worthwhile, since it's not unreasonable for the system described to be proportionately more effective against lower-energy shots. (the higher-energy ones would do more damage to the crystal matrix itself)

Reveraine (approved):

PR #530, opened 2018-06-16
Rework Long Rifle Expertise

Turtlelord26:
Some clarity updates/suggestions in previous skill substitutions.

Long Rifles, by their nature, live at the longest ranges and are at distinct disadvantages at any other due to their highly specialized designs. They also tend to be slower-firing and highly armor piercing, indicating a single-target focus.
0: Shoot even farther. It's what makes these guns special.
12: Another option is to replace Stealth checks to remain unobserved while motionless, but I went with Athl instead because I wanted more diversity in skill replacements and it's not necessarily the case that all snipers want to be stealthy. I figured it was sane that special sniper training includes getting to good vantage points.
24: Sneakerly and well-trained in the art of not getting shot back at.
36: Hint, hint, hint. Single target damage. Also encourages a good vantage point and somewhat alleviates range-based acc penalties.
48: Biggest weakness of long rifles is when someone walks right up to you. This doesn't actually make that a remotely favorable situation, but it does make it significantly less bad.
60: Inspired by Tech Rounds but balanced without them to try to avoid this being a mandatory feat for marksmen. Snipers should value having the right tool for the right job, especially as they should have to most time to swap tools of any class.

### DIFF
--- a/CharacterCreation/03_Feats.txt
+++ b/CharacterCreation/03_Feats.txt
@@ -697,7 +697,7 @@ have no prerequisites.
 	single target without losing line of sight to it.
 48:	Halve any minimum range associated with your long rifle and attachments, to a max reduction 
 	of 5m.
-60:	+10% Base damage.Increase by 50% any numerical benefits of specialty ammunition you fire 
+60:	+10% Base damage. Increase by 50% any numerical benefits of specialty ammunition you fire 
 	(does not affect Marksman Tech Rounds feats).
 	
 == Heavy Weapon Expertise ==
--- a/CharacterCreation/03_Feats.txt
+++ b/CharacterCreation/03_Feats.txt
@@ -664,7 +664,7 @@ have no prerequisites.
 36:	+1 Accuracy to ranges R1 and R2 while firing while benefiting from cover of any kind
 	against your target.
 48:	+10m to range R1, +5m to range R3
-60:	+10% Base damage; For up to one Action per round you may fire your carbine one-handed with  
+60:	+10% Base damage. For up to one Action per round you may fire your carbine one-handed with  
 	a -2 Accuracy penalty and double any automatic fire Accuracy penalty. Any foregrips or 
 	other attachments requiring the second hand do not provide their benefits during the 
 	Action, but your offhand is free to do something else. If you have taken aim at a target, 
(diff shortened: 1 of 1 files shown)


CharacterCreation/03_Feats.txt
```diff
@@ -675,26 +675,30 @@ have no prerequisites.
...
+12:	May use Weapons - Long Rifle (STR) in place of Athletics in order to get to hard-to-reach
+	places, like towers or treetops, that can serve as sniper nests.
+24:	While prone, benefiting from at least light cover, and able to see at least one enemy, 
+	gain an additional +1 cover bonus.
+36:	Gain a stacking +0.5 Accuracy each time you consecutively spend an action shooting a 
+	single target without losing line of sight to it.
+48:	Halve any minimum range associated with your long rifle and attachments, to a max reduction 
+	of 5m.
+60:	+10% Base damage. Increase by 50% any numerical benefits of specialty ammunition you fire 
+	(does not affect Marksman Tech Rounds feats).
```
bleehu, line 708: Quoth Rev "I really like those! They feel very sniper-y!"

CharacterCreation/03_Feats.txt
```diff
@@ -675,26 +675,30 @@ have no prerequisites.
...
-60:	+10% Base damage; Once per turn may change aim between targets within 30 degrees 
-	as a free action by passing a DC 75 dexterity check
+0:	+10% to range R3.
+12:	May use Weapons - Long Rifle (STR) in place of Athletics in order to get to hard-to-reach
+	places, like towers or treetops, that can serve as sniper nests.
+24:	While prone, benefiting from at least light cover, and able to see at least one enemy, 
+	gain an additional +1 cover bonus.
+36:	Gain a stacking +0.5 Accuracy each time you consecutively spend an action shooting a 
+	single target without losing line of sight to it.
+48:	Halve any minimum range associated with your long rifle and attachments, to a max reduction
```
bleehu, line 704: I bemoan a little bit the removal of "wrath of a patient man", which I wrote, but I like the pattern that you're keeping.

bleehu (approved):
I love it! Great work, Turtle!

CharacterCreation/03_Feats.txt
Turtlelord26: Honestly, it could work as its own standalone feat. As far as I'm concerned, feel free to reintroduce it as such!

PR #532, opened 2018-06-18
Rework Explosives Expertise

Turtlelord26:
Explosives includes hand grenades, grenade launchers, rocket launchers, artillery (including non-automatic Engineer artillery emplacements), and emplaced explosives like C10.

...Bleehu. Why did I not see this before. Did you intentionally name our C4 clone C10 because it translates as a Roman Numeral to CX?

Explosives are loud, obnoxious, area-of-effect weapons. Those with magazines tend to have limited capacity, and the ammo itself is quite large, limiting the amount that can be carried into battle. They aren't quiet and they don't last forever, but they hit like trucks, damage multiple enemies, and have some deceptively powerful utility.

0: It's something of a trend now for me to give the 0 bonus to R2 if I don't find a strong reason to put it anywhere in particular.
12: The Classic.
24: As yet, this mechanic has only been referenced by DM exposition on high explosives rolls. I am effectively proposing to turn it into a real mechanic specifically to counter the grenade-kick feat (naturally, with some risk. The cooking mechanic should be fleshed out in combat rules). As a corollary, without this specialized training, trying to cook a grenade is like ... trying to flail around untrained with a lightsaber. The anti-tampering device bit implies a suite of explosives enhancement item that we don't have but maybe want. I am thinking of putting the baseline critical failure chance in the realm of 5%, with ... predictable consequences.
36: This line also implies an as-yet unwritten mechanic for indirect fire. I'll Issue both if this is approved.
48: Extends the range at which you get maximum accuracy to include more space where you don't blow yourself up too.
60: This doesn't feel very capstoney to me, but the power level is undeniable. If someone else has a flashy idea for an explosives capstone please poke me with it.

Oh, and concerning the original grenade capstone - it's just such a basic rule of using explosives that I want to turn it into a baseline mechanic written directly into fundamental grenade rules, and trust player RP to use it or not based on individual character wisdom and experience. Also to be Issued on approval.

### DIFF
--- a/CharacterCreation/03_Feats.txt
+++ b/CharacterCreation/03_Feats.txt
@@ -732,12 +732,18 @@ have no prerequisites.
 grenade launchers, rocket launchers, and artillery, but not to explosive ammo of 
 other weapons or class abilities. Does apply to Engineer mortars.
 
-0:	+1 Accuracy with thrown explosives			
-12:	-10% reload miss chance (grenade launchers, rocket launchers, etc.)			
-24:	+5m to all range brackets for explosive launchers (grenade launchers, etc.)			
-36:	-0.5 miss chance @ all ranges (grenades, rocket launchers, etc.)			
-48:	+10% to Cha-based Intimidation checks when holding any explosive/launcher			
-60:	+15% Base damage, Allies in your blast radii may make a DC 60 Reflex save to take half damage (before armor)
+0:	+10m to range R2. If this would cause R2 to be greater than R3, 
+	also increase R3 to match the new R2.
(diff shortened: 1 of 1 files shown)


bleehu (approved):
Suggestion; just call the given capstone something cool like, "shock and awe". Maybe grant will saves to explosives? Nonblocking.

Turtlelord26:
Naming a line item would be inconsistent with how I handled the rest of them.
I think adding an additional save would be a unnecessarily complicated way of inflicting disadvantage...
...
...
Disadvantage is much cleaner than +50 DC. brb.

bleehu (approved):
Shipit!

PR #535, opened 2018-06-18
Reintroduce Extra-crits

Turtlelord26:
1Source removed them because he was concerned about the extra die roll necessary to distinguish crits from extra-crits. In dev conversation we were of the opinion that that particular extra roll is a fun extra roll and decided we wanted to keep extra-crits. Here it is accomplished.


bleehu (approved):
Shipit! We can close the issue in the design repo. We need to add Critical and Extra Critical to the glossary.

Issue #474, opened 2018-04-20
Update Character Creation README.txt


bleehu:
No really, it's embarassingly old and unhelpful. Let's nuke it and replace it with something helpful before it confuses someone.

PR #536, opened 2018-06-18
replaced the old character readme 

bleehu:
with one that doesn't have outdated rules. Issue #474 

 Changes to be committed:
	deleted:    README.txt
	new file:   Readme.md


Turtlelord26 (approved):
I like it. Short and simple.

Issue #499, opened 2018-05-03
Pull rules out of item docs


bleehu:
To comply with our new guideline of not putting rules in more than one document and instead standardizing in rules documents.
https://github.com/bleehu/CX_Design_Guide/tree/master/DevProcesses#guidelines

PR #541, opened 2018-06-18
Moved a bunch of rules from throwing weapons

bleehu:
to combat rules. Also did some minor restyling to combat rules.

Part of Issue #499 but not all of it.

 Changes to be committed:
	modified:   01_02_Combat_Rules.txt


Turtlelord26 (approved):
Yes, excellent. Exactly what I wanted to see when I raised the original issue. 🐢 Good style sniping too. May need a writing pass later to mesh it more cleanly into its new home but that is not the scope of this issue.

bleehu:
There's already an issue in the make play playable project just for that!

PR #544, opened 2018-06-19
added section on Security Drone names

bleehu:
I'm also considering suggesting a history where the sentient security drones have an MLK/Black Panther civil unrest style event which in turn informs the name scheme difference as a part of species pride. Or is using oppressing machines to talk about oppressing minorities a little too on the nose/distasteful?


Reveraine (approved):
I like it!
Ship it!

PR #546, opened 2018-06-20
Added Melee Expertise and Improvisational Brawler

Turtlelord26:
Let's see about sneaking this in before expertise goes to dev.

> By default Melee expertise does not apply to improvised weapons, for flavor reasons. So I added 
> another feat to address that.

> Melee weapons are somewhat different from guns, and it is harder to follow the expertise design 
> guidelines here. I decided to focus on melee vs melee combat since the challenge of melee vs guns is 
> all about getting into range (only the 0 bonus really helps) and not remotely in anything you do to the 
> poor sniper after you're already in his/her face.
> 0:  First range buff replaced with a close equivalency in move speed. Guardrailed to prevent exploits 
> with switching to melee, moving, then back to a firearm.
> 12: Skill substitution is Medicine for healing wounds caused by melee/thrown weapons and similar 
> effects, reflecting that swordsmen know a lot more about slashing wounds than gunners.
> 24: Disarm resistance is vital to the successful melee combatant. Don't. let. go. of. the. sword.
> 36: Conditional accuracy slot reflects the advantage of training in melee combat
> 48: Second range modifier replaced with conditional Guard increase to complement the 36 slot. Guard 
> comes after Accuracy because defending is harder than attacking.
> 60: Capstone grants a stacking crit chance bonus in a way that emphasizes 'studying the opponent' 
> and 'looking for openings' tenets of melee combat. This needs to be watched to make sure it doesn't 
> enable a horribly broken crit build, but I tried to make it as safe as possible by requiring the ramping-
> up time.

> Once I look into throwing/bow expertise, this unboarded project should be concluded.

### DIFF
--- a/CharacterCreation/03_Feats.txt
+++ b/CharacterCreation/03_Feats.txt
@@ -246,6 +246,13 @@ Prerequisites:
 ranged weapons. In taking this feat you sacrifice the ability to use PER on 
 rolls dealing with using a weapon.
 
+== Improvisational Brawler ==
+Prerequisites:
+*  Melee Expertise
+
+	The benefits of the Melee Expertise feat now also apply to Improvised Melee 
+Weapons.
+
 == Improvise Weapon ==
 Prerequisites:
 
@@ -838,7 +845,25 @@ other weapons or class abilities. Does apply to Engineer mortars.
 	also set R2 equal to the new R1.
 60:	+10% Base damage. Other characters make Shock, Reflex, and Will saves associated 
 	with your explosives with Disadvantage.
-		
+
+== Melee Expertise ==
+
+	Applies to Melee Weapons that the user has Proficiency with and Unarmed Strike. 
+Does not apply to Improvised weapons. Reminder: these bonuses are only in effect while 
+you are wielding an appropriate weapon.
+
+0:	+1 Move Speed while using the Charge action with intent to engage an enemy in 
+	Melee Combat.
+12:	May use Weapons - Melee (Int) in place of Medicine on skill checks related to 
+	triaging and healing wounds inflicted by melee weapons, nonexplosive throwing 
+	weapons, or environmental damage that approximates a melee strike.
+24:	When defending against a disarm attempt, roll maneuver checks with Advantage.
+36:	+1 Accuracy when engaging a target wielding a melee weapon (unarmed counts).
+48:	+1 Guard when engaging a target wielding a melee weapon (unarmed counts).
+60:	+10% Base damage. Gain +0.5 Critical Hit range whenever you hit a target but
+	do not score a Critical Hit. This bonus resets if you score a Critical Hit, 
+	attack a different target, are stunned, or leave Melee Combat with that target.
+
 ==============================
 Active Feats
 ==============================
--- a/CharacterCreation/03_Feats.txt
+++ b/CharacterCreation/03_Feats.txt
@@ -709,7 +709,8 @@ them all. The feats are formatted as a list of "X: Benefit", where X is the numb
 points required in the corresponding Weapon skill to gain the benefit.
 	
 	Weapon Expertise feats require proficiency with the weapon type, but otherwise 
-have no prerequisites.
+have no prerequisites. Note that as all characters are proficient with Unarmed Strike, 
+Melee Expertise effectively has no prerequisite.
 
 == Pistol Expertise ==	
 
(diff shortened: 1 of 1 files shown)


CharacterCreation/03_Feats.txt
```diff
@@ -246,6 +246,13 @@ Prerequisites:
 ranged weapons. In taking this feat you sacrifice the ability to use PER on 
 rolls dealing with using a weapon.
 
+== Improvisational Brawler ==
+Prerequisites:
+*  Melee Expertise
+
+	The benefits of the Melee Expertise feat now also apply to Improvised Melee 
+Weapons.
```
bleehu, line 254: I like it! I can think of a couple of players who will totally take this!

CharacterCreation/03_Feats.txt
```diff
@@ -838,7 +845,25 @@ other weapons or class abilities. Does apply to Engineer mortars.
...
+
+	Applies to Melee Weapons that the user has Proficiency with and Unarmed Strike. 
+Does not apply to Improvised weapons. Reminder: these bonuses are only in effect while 
+you are wielding an appropriate weapon.
+
+0:	+1 Move Speed while using the Charge action with intent to engage an enemy in 
+	Melee Combat.
+12:	May use Weapons - Melee (Int) in place of Medicine on skill checks related to 
+	triaging and healing wounds inflicted by melee weapons, nonexplosive throwing 
+	weapons, or environmental damage that approximates a melee strike.
```
bleehu: Blocking: This seems like quite a stretch. Being good at putting slices into characters does not translate into sewing them back up. As a suggested replacement, I notice that there isn't a bonus here for breaking out of a grapple, which is one of the first things almost all martial arts teaches.

CharacterCreation/03_Feats.txt
```diff
@@ -838,7 +845,25 @@ other weapons or class abilities. Does apply to Engineer mortars.
...
+24:	When defending against a disarm attempt, roll maneuver checks with Advantage.
+36:	+1 Accuracy when engaging a target wielding a melee weapon (unarmed counts).
+48:	+1 Guard when engaging a target wielding a melee weapon (unarmed counts).
+60:	+10% Base damage. Gain +0.5 Critical Hit range whenever you hit a target but
+	do not score a Critical Hit. This bonus resets if you score a Critical Hit, 
+	attack a different target, are stunned, or leave Melee Combat with that target.
```
bleehu: Does killing a target count as leaving melee combat? If it does, I think this ultimate is underpowered. Most enemies that aren't also melee spec'd tend to die in a couple of melee hits. That cadence is all but necessitated to maintain the viability of melee in the squad shooter paradigm. That said, 10% more damage makes this only a little underpowered.

bleehu (changes requested):
There's some really good stuff here that is making our couple of melee-centered classes much more viable. I've requested a couple of tweaks. 

I almost want to suggest breaking melee feats up into light and heavy; I think that would go a long way into giving them more "personality"; giving light weapons a chance to reposte and giving heavy weapons a chance to break through gaurd by sheer force, etc. But I understand that it clear scope creep.

CharacterCreation/03_Feats.txt
Turtlelord26: The base damage buff can almost be factored out of the power analysis (relative to other expertise feats)  since it's on all of the 60 point bonuses. 
I had forgotten how few hits we've designed characters to take, and am quite willing to test with keeping the bonus past killing the target. Do you think it would be appropriate to also change 'leave Melee Combat' to 'spend 1 round not in Melee Combat' to allow carrying the bonus between nonadjacent enemies?

CharacterCreation/03_Feats.txt
Turtlelord26: An excellent call, will nix existing skill replacement.

Turtlelord26:
One precedent I set with the Explosive Expertise is putting two effects on one bonus aimed at different subgroups of the weapon category. I could easily see one of the bonuses giving both of your suggested category-specific bonuses, the only question would be what I replace.

CharacterCreation/03_Feats.txt
bleehu: I like how that gives a sort of "rampage" feeling to it, but I'm concerned that 1, that rampage feeling is already conferred by things like Flash Dance, and 2, that critical hits aren't what make melee weapons unique. Instead, I'm asking myself what is unique about melee weapons? Well, that you have to be next to your target to hit them... What if instead, the ultimate skill would be to be in partial cover using whomever you're in melee combat with as long as you're engaged in melee combat? Perhaps even forcing an enemy attempting to fire at you to make an explicit Luck roll when shooting at you, risking that they might hit their friend instead?

bleehu: The obvious drawback of my proposed alternative is that it doesn't find a way of keeping the good work you've done, and I'm loathe to throw away good design.

Turtlelord26: Don't we already have a mechanic with Luck rolls when shooting into a melee, or have you been houseruling that?

bleehu: Totally been house ruling it

Turtlelord26: @.@ Ok lemme fix that because that should be baseline. You don't have to be all that good at melee to mess up other people's firing solutions from 30m away.

Turtlelord26:
There have been changes. Capstone is still on the discussion table, as are the new 12 point and 24 point bonuses (and of course anything else anyone wants).

CharacterCreation/03_Feats.txt
```diff
@@ -838,7 +846,32 @@ other weapons or class abilities. Does apply to Engineer mortars.
...
+== Melee Expertise ==
+
+	Applies to Melee Weapons that the user has Proficiency with and Unarmed Strike. 
+Does not apply to Improvised weapons. Reminder: these bonuses are only in effect while 
+you are wielding an appropriate weapon.
+
+0:	+1 Move Speed while using the Charge action with intent to engage an enemy in 
+	Melee Combat.
+12:	May use Weapons - Melee (Str/Dex) in place of Acrobatics or Athletics on 
+	maneuver checks to avoid, gain control of, or escape a Grapple; defend against
```
bleehu, line 859: love it!

CharacterCreation/03_Feats.txt
```diff
@@ -838,7 +846,32 @@ other weapons or class abilities. Does apply to Engineer mortars.
...
+	maneuver checks to avoid, gain control of, or escape a Grapple; defend against 
+	Disarm attempts; or defend against Trip attempts.
+24:	While wielding a Light or One-handed non-Hybrid weapon, whenever an enemy attacks 
+	you in Melee Combat and ties your Guard, you may make a melee attack against that 
+	enemy as a Reaction. 
+	While wielding a Hybrid or Two-handed weapon, you may choose to gain +2 to melee 
+	Accuracy checks for one round, but if you do then until the start of your next 
+	turn you take -2 to Guard checks.
+	If dual-wielding a Hybrid and Light or One-handed weapon, you may only use one 
+	of these effects per round, but may choose which.
```
bleehu, line 868: love it!

bleehu (approved):
Still a little underwhelmed by the melee capstone, but I think it's still quite viable as a feat!

PR #548, opened 2018-06-20
added IR as a trait for Security Drones

bleehu:
I think having the Security Drone's perk of having IR goggles makes a ton of sense as an available trait. What do you think?


Turtlelord26 (approved):
V nice 👍 

More powerful that just saving a couple hundred $ in IR goggles because they can't be disarmed (at least, without making your lack of IR the least of your problems)

Reveraine (approved):
Love it!

PR #549, opened 2018-06-20
added the Series ten Cinder Drone

bleehu:
I had origionally intended this as a large or heavy series, but got carried away with more flame-retardant type robots. What do you think?


Turtlelord26 (approved):
I like both this idea and the implication that cold-environment drones also exist.

Nonblocking spellcheck on Opaleite in line 115.

Reveraine (approved):
Yes! Good! Ship It!

Issue #557, opened 2018-06-23
Gripe: Combat stat bonus table is in the wrong place.


dfuchs2:

bleehu:
Specifically, the combat stat bonus table was expected by players to be in the combat rules document. https://github.com/bleehu/Compound_X/blob/dev/01_01_Basic_Rules.txt#L249

PR #568, opened 2018-06-25
Turtlelord26 #557 cmb stat table >cmb rules

Turtlelord26:
Moved the table and associated rules text block. Some editing in the first paragraph to reflect the new location and some rules updates that had not yet propagated thereto.
From #557


bleehu (approved):
Shipit! Much easier to find!

Issue #558, opened 2018-06-23
Gripe: Hitting what you intend outdated terminology


Reveraine:
References miss chance, should reference Guard (non parenthetically). Nothing else about it needs to change.
-Turtle

bleehu:
https://github.com/bleehu/Compound_X/blob/dev/01_02_Combat_Rules.txt#L86

PR #569, opened 2018-06-25
Update Mechanic: Hitting what you intend

Turtlelord26:
#558 
Now references Guard and not miss chance.


bleehu (approved):
much better! Shipit!

Issue #559, opened 2018-06-23
Gripe: Need reaction check


dfuchs2:

bleehu:
Or do we want characters that have reaction abilities to be able to react to more than 12 things per round?

Turtlelord26:
I'm open to the possibilities of feats or abilities expanding character's abilities to react, but I want a baseline limit of 1/round.
If we start messing with that, remember that some of the most broken content I've seen out of DnD has involved attacks as Reaction-equivalent actions with high or unbounded limits on the number available.

bleehu:
Can we expand that cap to three per round? Most other reactions also have another cost to them, or at least they should. I want pre-emptive strike to be the bane of the Antlion Hatchling rush, but not be broke-as-hell. I think that balance is closer to a 3 per round than 1 per round.

Turtlelord26:
We can test that. Fair warning that this increases the likelihood that something in the future that looks ok breaks stuff. Would like a design note for that extra cost to Reactions.

bleehu:
https://github.com/bleehu/CX_Design_Guide/issues/102

PR #571, opened 2018-06-26
Turtlelord26 reactions and initiative

Turtlelord26:
#559 
With lots and lots of scope creep.

Reactions now limited to 1 per round, Initiative now actually has rules, and Free Actions have a rules definition. I didn't check the glossary.

### DIFF
--- a/01_02_Combat_Rules.txt
+++ b/01_02_Combat_Rules.txt
@@ -6,8 +6,8 @@ Combat
 	Combat starts when either you or an enemy decides they would like to make an 
 attack (or as the GM decrees). At that point, all involved characters roll 
 Initiative to determine the order in which they take turns. Initiative is a d100 roll 
-that is modified only by Soldier class passive abilities. The section "Actions", 
-below, has more detail on turns in combat.
+that is modified only by Soldier class passive abilities, and higher numbers go 
+first. The section "Actions", below, has more detail on turns in combat.
 
 ===== Combat Stat Modifiers for Ranged and Melee Accuracy =====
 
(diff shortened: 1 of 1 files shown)


Reveraine (approved):
Excellent! Ship it!

01_02_Combat_Rules.txt
```diff
@@ -383,20 +385,32 @@ Performing a Skill Check (other than Getting Bearings): 1 Action
...
-below) or Opportunity Attacks (described in "Melee Combat", above). After the 
-reaction occurs, play continues as normal (if it is an opponents turn, they 
-continue their turn). 
+Reactions:
+	Certain abilities or circumstances will allow you do something 
+in response to a triggering event. Doing so requires using a special kind 
+of Action called a Reaction. Reactions occur at the immediate moment 
+that they are triggered, regardless of whose turn it is. Common uses of 
+Reactions include Opportunity Attacks (described in "Melee Combat", above) 
+and Held Actions, below. After the Reaction occurs, play continues as normal.
```
rosebrallier, line 404: Do we really want held actions to be a type of Reaction? I'm wary of discouraging held actions like that, and I like the tactical depth allowed when you can both take a reaction (based on a feat, ability, opportunity attack) and shift part of your turn within the initiative order.

rosebrallier (changes requested):
GLORIOUS
I do have some concerns with taking a held action being mutually exclusive with taking any other kind of reaction.

01_02_Combat_Rules.txt
Turtlelord26: While this problem was technically created by these changes, the actual problem text isn't something I touched. I'm not sure where I fall on this question (therefore it is a good question 🐢), but since I've scope crept so much already would you be comfortable classifying this as New Issue instead of Blocker?

rosebrallier: Aye aye!

rosebrallier (approved):

bleehu (approved):
shipit!

01_02_Combat_Rules.txt
Turtlelord26: Now that the Reaction cap has moved to 3, has this become a nonissue? @rosebrallier

Issue #506, opened 2018-05-14
Gripe: Using medkits/repairkits to Stabilize


Turtlelord26:
Currently there are no listed mechanics for using medkits or repair kits to enhance attempts to stabilize downed characters. Can has plox?

rosebrallier:
Done?

Turtlelord26:
Thank for catch

PR #590, opened 2018-07-06
Clarify use of Repair/Medkits to Stabilize

Turtlelord26:
#506
Now provides +20 to the check. Also did some clarity work on the Stabilization section.

### DIFF
--- a/01_02_Combat_Rules.txt
+++ b/01_02_Combat_Rules.txt
@@ -899,7 +899,7 @@ of Stabilization. Note that most persistent damage effects, like bleeds or burns
 fall off of characters when they are Downed.
 
 	If a character fails a stabilization check, nothing happens. It's just 
-taking longer to stem the bleeding. On critical success (91-100 before modifiers), 
+taking longer to stem the bleeding. On critical success (base Critical range 10), 
 your excellent care allows you to choose whether the target counts this revival 
 against their revivals per encounter limit. On critical failure (01-05 before 
 modifiers), the bleeding (or equivalent) gets worse: the target must immediately 
--- a/01_02_Combat_Rules.txt
+++ b/01_02_Combat_Rules.txt
@@ -889,17 +889,30 @@ health. Any additional attacks you sustain will push you into negative health.
 
 Stabilize:
 	As an action, a character may choose an adjacent, Downed character and 
-attempt to Stabilize them. If successful on a DC 80 Medicine (Dex usually) check, 
-the character now has 1 HP, is no longer Downed, and may take Actions as normal 
-within the initiative order. Note that most persistent damage effects, like 
-bleed or burn tokens, fall off of characters when they are Downed.
+attempt to Stabilize them. That character makes DC 80 Dex-based skill check: 
+Medicine for primarily organic targets, or Repair for primarily mechanical targets.
+If successful, the target is revived and now has 1 HP, is no longer Downed, and 
+may take Actions as normal within the initiative order. A character may be revived 
+in a single encounter a number of times equal to half their Fort, minimum two. Past 
+this limit, stabilized characters are unconscious in addition to the other effects 
+of Stabilization. Note that most persistent damage effects, like bleeds or burns, 
+fall off of characters when they are Downed.
+
+	If a character fails a stabilization check, nothing happens. It's just 
+taking longer to stem the bleeding. On critical success (91-100 before modifiers), 
+your excellent care allows you to choose whether the target counts this revival 
+against their revivals per encounter limit. On critical failure (01-05 before 
+modifiers), the bleeding (or equivalent) gets worse: the target must immediately 
+roll an additional Death Save. If the total check is then still high enough to 
+succeed and the target is still alive, they are then Stabilized as normal.
 
 	Up to one character may assist with a stabilization check. Either involved 
 character may expend one Medkit (for primarily organic targets) or one Repair Kit 
 (for primarily mechanical targets) to grant +20 to the Medicine check. Feel free 
-to scavenge one from the target, they probably won't make a fuss.
+to scavenge one from the target, they probably won't make a fuss. One kit provides 
+a bonus to all Stabilization attempts made on the same target in one round.
 
-	Don't forget that a recently stabilized character may need  to take actions to
+	Don't forget that a recently stabilized character may need to take actions to
 retrieve a dropped weapon or stand back up if they have fallen.
 
 *Some hazards can reduce you to zero health points and kill you 
--- a/01_02_Combat_Rules.txt
+++ b/01_02_Combat_Rules.txt
@@ -888,10 +888,16 @@ above. Even if you had 5 health and took 80 damage, when you go down you go to 0
 health. Any additional attacks you sustain will push you into negative health. 
 
 Stabilize:
-	As an action, a teammate may bring you back up to 1 health point by succeeding 
-on a DC 80 Medicine skill check to stabilize you. Medicine skill, INT mod, and any
-bonuses from medical equipment such as medic-kits are added to the roll. Once stabilized,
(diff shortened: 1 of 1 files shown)


01_02_Combat_Rules.txt
```diff
@@ -888,10 +888,16 @@ above. Even if you had 5 health and took 80 damage, when you go down you go to 0
 health. Any additional attacks you sustain will push you into negative health. 
 
 Stabilize:
-	As an action, a teammate may bring you back up to 1 health point by succeeding 
-on a DC 80 Medicine skill check to stabilize you. Medicine skill, INT mod, and any
-bonuses from medical equipment such as medic-kits are added to the roll. Once stabilized,
-you may take actions  and speak as normal.
+	As an action, a character may choose an adjacent, Downed character and 
+attempt to Stabilize them. If successful on a DC 80 Medicine (Dex usually) check,
```
rosebrallier: "usually Dex" reason: formal tone

01_02_Combat_Rules.txt
```diff
@@ -888,10 +888,16 @@ above. Even if you had 5 health and took 80 damage, when you go down you go to 0
...
 Stabilize:
-	As an action, a teammate may bring you back up to 1 health point by succeeding 
-on a DC 80 Medicine skill check to stabilize you. Medicine skill, INT mod, and any
-bonuses from medical equipment such as medic-kits are added to the roll. Once stabilized,
-you may take actions  and speak as normal.
+	As an action, a character may choose an adjacent, Downed character and 
+attempt to Stabilize them. If successful on a DC 80 Medicine (Dex usually) check, 
+the character now has 1 HP, is no longer Downed, and may take Actions as normal 
+within the initiative order. Note that most persistent damage effects, like 
+bleed or burn tokens, fall off of characters when they are Downed.
```
rosebrallier: Are we consistently calling them "tokens"?

rosebrallier (changes requested):
When we stabilize a mechanical target, should it be Medicine or Repair check?
And what happens on a failed check?

01_02_Combat_Rules.txt
```diff
@@ -888,12 +888,31 @@ above. Even if you had 5 health and took 80 damage, when you go down you go to 0
 health. Any additional attacks you sustain will push you into negative health. 
 
 Stabilize:
-	As an action, a teammate may bring you back up to 1 health point by succeeding 
-on a DC 80 Medicine skill check to stabilize you. Medicine skill, INT mod, and any
-bonuses from medical equipment such as medic-kits are added to the roll. Once stabilized,
-you may take actions  and speak as normal.
-
-	Don't forget that a recently stabilized character may need  to take actions to
+	As an action, a character may choose an adjacent, Downed character and
```
bleehu, line 891: Nonblocking; I think we should lengthen the time it takes to stabilize a character so that it's longer than the time to raise and fire a weapon. Perhaps I should also lengthen the time for the stabilize Doctor Procedure.

01_02_Combat_Rules.txt
```diff
@@ -888,12 +888,31 @@ above. Even if you had 5 health and took 80 damage, when you go down you go to 0
...
+of Stabilization. Note that most persistent damage effects, like bleeds or burns, 
+fall off of characters when they are Downed.
+
+	If a character fails a stabilization check, nothing happens. It's just 
+taking longer to stem the bleeding. On critical success (91-100 before modifiers), 
+your excellent care allows you to choose whether the target counts this revival 
+against their revivals per encounter limit. On critical failure (01-05 before 
+modifiers), the bleeding (or equivalent) gets worse: the target must immediately 
+roll an additional Death Save. If the total check is then still high enough to 
+succeed and the target is still alive, they are then Stabilized as normal.
```
bleehu, line 907: I love the crit-success/fail detail! Fantastic!

01_02_Combat_Rules.txt
```diff
@@ -888,12 +888,31 @@ above. Even if you had 5 health and took 80 damage, when you go down you go to 0
...
+
+	Up to one character may assist with a stabilization check. Either involved 
+character may expend one Medkit (for primarily organic targets) or one Repair Kit 
+(for primarily mechanical targets) to grant +20 to the Medicine check. Feel free 
+to scavenge one from the target, they probably won't make a fuss. One kit provides 
+a bonus to all Stabilization attempts made on the same target in one round.
```
bleehu, line 913: I also really really like how you called out assistance as a specific mechanic! Great work!

bleehu (approved):
One nonblocking issue, but we can handle that elsewhere.

01_02_Combat_Rules.txt
Turtlelord26: Though I didn't mention what it does. We'll need a general rule elsewhere. *New Issue

01_02_Combat_Rules.txt
Turtlelord26: I concur, but if we change one we should change the other too, and that makes it scope creep I think?
If we decide on longer than 2 actions, we can also delete the bit I put in here about reusing repair/medkits.

01_02_Combat_Rules.txt
```diff
@@ -888,12 +888,31 @@ above. Even if you had 5 health and took 80 damage, when you go down you go to 0
...
+Medicine for primarily organic targets, or Repair for primarily mechanical targets.
+If successful, the target is revived and now has 1 HP, is no longer Downed, and 
+may take Actions as normal within the initiative order. A character may be revived 
+in a single encounter a number of times equal to half their Fort, minimum two. Past 
+this limit, stabilized characters are unconscious in addition to the other effects 
+of Stabilization. Note that most persistent damage effects, like bleeds or burns, 
+fall off of characters when they are Downed.
+
+	If a character fails a stabilization check, nothing happens. It's just 
+taking longer to stem the bleeding. On critical success (91-100 before modifiers),
```
rosebrallier: Did we kill Luck-based crit-range modification?

01_02_Combat_Rules.txt
```diff
@@ -888,12 +888,31 @@ above. Even if you had 5 health and took 80 damage, when you go down you go to 0
...
+this limit, stabilized characters are unconscious in addition to the other effects 
+of Stabilization. Note that most persistent damage effects, like bleeds or burns, 
+fall off of characters when they are Downed.
+
+	If a character fails a stabilization check, nothing happens. It's just 
+taking longer to stem the bleeding. On critical success (91-100 before modifiers), 
+your excellent care allows you to choose whether the target counts this revival 
+against their revivals per encounter limit. On critical failure (01-05 before 
+modifiers), the bleeding (or equivalent) gets worse: the target must immediately 
+roll an additional Death Save. If the total check is then still high enough to
```
rosebrallier: It's a little confusing to me that you could succeed at a task when you critically failed the roll.

rosebrallier (commented):
So much detail and specificity! So much good! I do have some concerns about critical success and failure.

01_02_Combat_Rules.txt
Turtlelord26: Negative, will fix. Note that the fix will reference currently nonexistent rules re critical success on skill checks (issued elsewhere).

01_02_Combat_Rules.txt
Turtlelord26: I was thinking success-at-a-cost style for this action, because I didn't think I would like ever auto-failing a non-Doctor serious enough to put 50+ skill points into Medicine (/non-Engineer with 50 points in Repair). Not unwilling to revert.

Turtlelord26: @bleehu your thoughts on this also solicited

rosebrallier (approved):

PR #625, opened 2018-08-01
Constructor Drone First Trait

dfuchs2:


bleehu:
Don't forget to put a quick synopsis of the changes you made in the commit log, that'll help @Turtlelord26 make change logs for when your work ships to RC and Play later!

bleehu (approved):
I like it! I kinda miss how security drones aren't big and bulky as a species anymore, and I'm glad you put that back in as at least an option!

Turtlelord26 (commented):
This trait looks like two traits stapled together: the attribute modifiers and the skill modifiers. I would wonder if we could separate them (and maybe swap the flat skill bonus to an Advantage roll).
Design question @Reveraine: Are attribute modifiers in species traits a thing we can/want to do?

Reveraine:
To write down what I said in stand-up, so we do not forget, attribute modifiers are unique to sub-species rather than actual traits.

dfuchs2:
Should I keep it as it is, or should I change this trait up a bit? I'm open to changing it if it makes it fit the current design better.

Turtlelord26:
The only change we require is that the attribute modifiers leave the trait and become subspecies passives (see other species files for format).

I think the rest of it is a good trait as it stands. I would personally consider swapping out the +20 for Advantage on those rolls, but that's your call.

bleehu:
Linking Issue #554

Issue #555, opened 2018-06-22
Create Second Avialable Trait for the Third Drone Series


bleehu:
Once we've added the new third Class-A Drone Series, we'll need to give them another available trait. Having more than one trait gives the players a meaningful choice. Please keep it balanced and hopefully work in some good differences in kind. https://github.com/bleehu/CX_Design_Guide/blob/master/DevProcesses/welcome.md#differences-in-scale-and-differences-in-kind
Be creative!

While yes, there are multiple issues on the same document, please refrain from hitting many birds with one Pull Request; If there's a problem with one Available Trait and not the other, we don't want one to be stuck in review and blocking the other. The traits should be independent of each other; they are one atomic change. Thus they should be PR'd and considered independently.

Blocked by #553

bleehu:
Handled in PR #626

PR #626, opened 2018-08-01
Constructor Drone Second Trait

dfuchs2:


CharacterCreation/01_02_Species_ECSC_Security_Drone.txt
```diff
@@ -229,3 +229,11 @@ drones, to deal with the constant heavyweight lifting that is required of them.
     
 Size: Large
 Favored Terrain: Urban
+
+Engineering-Minded:
+    The Series Twelve Drone is programmed to better understand structural stability and 
+weaknesses. Knowledge is powerful, and Constructor Series Drones know exactly where to 
+apply force to do the most damage. Gain +20 when trying to weaken or demolish structures 
+such as walls or doors. Constructor Drones are built to be very resourceful when it comes
+to building. Gain Jury-Rigging at 15 if you take it as a level 1 skill, or at 20 if you
```
bleehu: 15 and 20 are a little too high for level 1 with @rosebrallier's new skill system, where 6 or 12 would be the max expected of a level 1 character. Also, Engineers already get a flat bonus of 10 to Jury-rigging as well as proficiency in the skill. I really like the idea that construction bots are better at learning to build things, so I hope we keep the theme. I really like it. Rev suggests changing this to advantage on jury-rigging instead?

bleehu (changes requested):
This is really good, but a bit OP. I think the obvious synergy between Construction bot as a species and engineer as a class has been telegraphed well by their names. But we'll need to nerf the bonus and/or spread the benefit around a bit because this turns into like a 30-something jury rigging at level 1, which is higher than we want.

Turtlelord26 (commented):
This one also looks like two traits. Most of the traits currently in place are fairly narrow, which allows more customization when building characters.

I would suggest (again) to address blee's first concern by swapping out the flat bonuses to Advantage rolls. Advantage is pretty good tech because it's less powerful than +20 but feels so much cooler. Advantage will also be more powerful in the late game because, unlike giving a bunch of points in jury-rig at level 1, it won't count against the 60 skill point cap.

bleehu (approved):
Shipit!

bleehu:
Linking Issue #555

PR #632, opened 2018-08-03
rearranged damage types and armor to surface fall damage rule, did so…

bleehu:
…me style cleanup

### DIFF
--- a/01_04_Damage_Types_and_Armor_Types.txt
+++ b/01_04_Damage_Types_and_Armor_Types.txt
@@ -4,6 +4,8 @@ Damage Types and Armor Types
 
 ===== Damage Types =====
 
+== Weapon Types ==
+
 Ballistic Weapons:
     Ballistic weapons are straightforward: they are blocked by armor and ignore 
 shields. Any damage that defeats armor damages health.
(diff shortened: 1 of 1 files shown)


01_04_Damage_Types_and_Armor_Types.txt
```diff
@@ -21,6 +23,24 @@ shielding. Lasers deal full damage to armored targets; If a player wearing
 ballistic armor is hit with a laser, they receive damage as if they weren't 
 wearing any armor.
 
+== Fall Damage ==
+
+    Beyond weapon damage, there are also environmental threats to consider. An 
+important note is on fall damage. A player can fall up to their Dex/2 in
```
Reveraine, line 29: Non-blocking: This wording is a little weird here. Something like:
Weapons are not the only danger in the universe; falling can be just as deadly.
Basically, I think the two sentences can be combined into something that flows more.

Reveraine (approved):
Other than a weird sentence structure, good!

01_04_Damage_Types_and_Armor_Types.txt
```diff
@@ -64,23 +85,12 @@ ignores armor.
 Armor:
     Ballistic Armors (or more simply, Armor) are the first type of protective 
 measure that you will come across. When armor is shot by a ballistic weapon, the 
-armor may reduce or fully block damage before the character's health is damaged. 
+armor may reduce or fully block damage before the character's health is damaged.
+Balistic armor also does not conduct electricity, and an electric weapon that
```
Turtlelord26, line 89: sp: "Balistic" -> Ballistic

Turtlelord26 (commented):
Fall damage seems more like a rule that I would find in Basic_Rules than it does a distinct type of damage.

If we're making Electric damage fizzle on hitting armor, we need to also say that in the Electric damage entry to avoid ambiguity.

PR #633, opened 2018-08-03
stylized and re-ordered skills document

bleehu:
Before:

before

After:
after


rosebrallier:
I'm not a fan of how the lists of skill subsets parse outside of the textbox and separate from their parent skill.
That said, this is a massive improvement over the previous iteration!

bleehu:
I did what I could with the limitations at hand. Unfortunately, the parser doesn't know how to nest one element in another.

rosebrallier (approved):
Shipit!

Issue #612, opened 2018-07-25
Update cloaking to not use nanites


bleehu:
as part of our push to update the nanite aesthetic to be less space magic, we need to change the armor document so that nanites no longer run cloaking and shield units.

PR #634, opened 2018-08-04
Turtlelord26 #612 nanites to energy

Turtlelord26:
I'm turning this into an explicit superbranch so that I have to PR my changes to it too, before my branch explodes with too many unreviewed changes.


rosebrallier (approved):
This is some truly excellent prose!

bleehu (approved):
shipit!

PR #636, opened 2018-08-06
Turtlelord26 Nanite Removal: Cloak Rules Update

Turtlelord26:
Nanites -> Energy in the Cloak Rules file. Also general wording improvement.

Removed Cloak Turns stat. Made new Overheat stat to cover that functionality - cloak duration is now based on your equipment instead of your level.
Clarified Charge Delay stat, massaged its numbers to make more sense with new Overheat.


bleehu (approved):
looks good! Shipit!

Issue #629, opened 2018-08-03
Gripe: Clarity on Overcharged Laser Attachment


Turtlelord26:
https://github.com/bleehu/Compound_X/blob/dev/Items/04_00_WeaponAttachments.txt#L530
> Jam Range: + 2 on second roll

Does this mean if I shoot 6 bullets, I roll 5 with Jam +0 and 1 with Jam +2 (literalist reading), or 1 with Jam +0 and 5 with Jam +2? Flavor would suggest the latter. 

Propose instead "Jam Range: +0, then +2 for each shot after the first. Resets whenever the weapon spends at least one Action not firing."
Though that is a little wordy. If anyone comes up with something shorter please feel free to use it.

bleehu:
My bad. I'll update it.

PR #639, opened 2018-08-06
clarified when overcharge ammo jams

bleehu:
Specifically to fix Issue #629


Turtlelord26 (approved):
Better, though I still see a small chance for misunderstanding. Maybe we need to same that second roll (Jam roll?) (Jam confirmation roll?), but that would be outside this scope.
Shipit

Issue #572, opened 2018-06-26
Heavy Weapon Expertise (36) and our Philosophy of Accuracy


Turtlelord26:
From the depths of #542 

> 36:	+1 Accuracy against targets when your weapon's APL is at least 1 greater         
> +	than the target's armor's APL.

  @bleehu
bleehu a day ago  Owner
I'm a little less ecstatic about this one. Why would you be more accurate against heavier armored targets? I have a feeling @rosebrallier would call aesthetic on it. I also notice that there's not a bonus to suppression on here. Could we instead do something like a bonus to accuracy rolls for shots taken in response to an enemy moving into suppression? My fixation on suppression in this case being the assumption that it's what heavy weapons are uniquely good at.

  @Turtlelord26
Turtlelord26 14 hours ago  Collaborator
Because Accuracy is not only whether your bullet finds the target, but whether the bullet actually does damage instead of careening off their armor in a glancing blow, I thought I could get away with this.
I did consider suppression, but the problem is that for medium and heavy suppression, there aren't any attack rolls involved, only automatic hits.

I have no objections to classifying it as a new issue if we're still not sure about this.

  @bleehu
bleehu 13 hours ago  Owner
Make it so!

  @rosebrallier
rosebrallier 38 minutes ago  Collaborator
I agree that we should make this a new issue; it brings up a broader problem of how we collectively want to imagine accuracy and hits.

rosebrallier:
Because our armor currently absorbs damage, I imagine a hit as impacting the target, whether it hits flesh or armor. It's odd for me to imagine a bullet glancing off armor without having any impact on that armor.

Turtlelord26:
True. But does that change if we go to a static armor system where Acc<x misses, x<=acc<=y hits armor, and y<Acc misses entirely?

EDIT: Acc<x misses ^armor^

rosebrallier:
Could you clarify the difference between "misses" and "misses entirely"? I gave it 24hrs and am still foggy on what you mean.
Also, will armor still absorb a set amount of damage/time when hit, or will that be exclusive to shields?

rosebrallier:
If armor only changes the DC to hit the target (as in d&d), then your perspective makes perfect sense to me. If armor absorbs damage in any way, then to me a miss cannot be described as a glancing blow.

Turtlelord26:
@.@ I'm unfavorably impressed by my imprecision. I meant misses ^armor^ for the first expression. 
Oops :(

[...]

PR #642, opened 2018-08-06
Rework Heavy Weapon Expertise 36 Pt Bonus

Turtlelord26:
Again.

#572 
Now gives accuracy bonus to the "Laying Down Fire" aesthetic.


bleehu (approved):
shipit!

rosebrallier (approved):
Ship-it!
Note: Right now, all Heavy Weapons have Med or Heavy suppression, so the bonus to suppression accuracy will only help future weapons with Light suppression.

PR #650, opened 2018-08-09
Version bump to 0.1.6

bleehu:
modified Readme.md


Turtlelord26 (approved):
Are you sure we've sufficiently tested this type of sweeping change? What if it breaks everything...

(Shipit!)

PR #653, opened 2018-08-13
Rewrote Critical Hit Rules

Reveraine:
There are now charts! And Terminology! And best of all! No Math!


Turtlelord26:
What do you mean no math D: D: D: D: D:
(starts reading the changes)

Turtlelord26 (changes requested):
Ok, this is actually pretty cool and probably a strict formatting improvement. 🐢 says good.

One formatting suggestion. In the second table, it might be easier to reference quickly if the first column is the Luck Stat again instead of the Base Crit Range, unless we plan to have a spot on the character sheet for both Base Crit Range and Actual Crit Range. We could also consider collapsing them into one larger table to avoid having to cross-reference at all.

I have a couple of questions about the mechanical details.
First, I'm wondering why as Luck increases the probabilities are weighting so heavily toward extra-crits (instead of the former half-half split) at the same time the crit range is going up. I think we reduce complexity by keeping the 50-50 split between crit types, though if there are other opinions I would be happy to discuss. Also, I would be in favor of letting the lowest luck bracket be split between non-crit and regular-crit (and no extra-crit possible) if we wanted to keep that unique.
Second, it's not currently clear whether a Luck of 10 is a +1 or a +2. I read 10+ as including 10.

Reveraine:
I changed things based on @Turtlelord26 suggestions =)

Turtlelord26 (approved):
Excellent 👍 
Shipit the thing!

bleehu (approved):
Shipit please!

01_02_Combat_Rules.txt
```diff
@@ -444,25 +444,35 @@ You can jump vertically Str/3 meters.
...
+automatic mode), you may roll high enough to use your Critical Hit Chance. 
+Critical Hit Chance is based on your Luck Stat, but can also be modified by 
+abilities, weapons, attachments, etc. 
+	If your Accuracy Roll is high enough, you activate your Critical Hit Chance.
+When this happens, you roll a second d10 to determine if you have performed a Critical 
+Hit, an Extra-Critical Hit, or, if you have less than average Luck, merely hit your enemy. 
+
+	A Critical Hit deals +20 damage, while an Extra-Critical Hit deals +30 damage. 
+
+	Some abilities or feats have additional effects that trigger when you score
```
rosebrallier: From a tone perspective, I'm not a great fan of "score." I suggest either using "perform" (as above), or changing both to "land" (which has combat connotations).

01_02_Combat_Rules.txt
```diff
@@ -444,25 +444,35 @@ You can jump vertically Str/3 meters.
...
-
-	You cannot score Critical Hits while firing a gun in burst or automatic mode.
+
+	When you perform an Accuracy Roll (while not firing a gun in burst or 
+automatic mode), you may roll high enough to use your Critical Hit Chance. 
+Critical Hit Chance is based on your Luck Stat, but can also be modified by 
+abilities, weapons, attachments, etc. 
+	If your Accuracy Roll is high enough, you activate your Critical Hit Chance.
+When this happens, you roll a second d10 to determine if you have performed a Critical 
+Hit, an Extra-Critical Hit, or, if you have less than average Luck, merely hit your enemy.
```
rosebrallier: The structure of this sentence is a bit wonky b/c the verb for the last clause ("hit") is different from the first two "performed" and the object of the last clause (your enemy) is inconsistent with the first two (a Crit Hit or Extra-Crit Hit).
Personally, I'd go with "if you have landed a Crit Hit, an Extra-Crit Hit, or,...Luck, merely a hit"

01_02_Combat_Rules.txt
```diff
@@ -444,25 +444,35 @@ You can jump vertically Str/3 meters.
...
+
+	When you perform an Accuracy Roll (while not firing a gun in burst or 
+automatic mode), you may roll high enough to use your Critical Hit Chance. 
+Critical Hit Chance is based on your Luck Stat, but can also be modified by 
+abilities, weapons, attachments, etc. 
+	If your Accuracy Roll is high enough, you activate your Critical Hit Chance.
+When this happens, you roll a second d10 to determine if you have performed a Critical 
+Hit, an Extra-Critical Hit, or, if you have less than average Luck, merely hit your enemy. 
+
+	A Critical Hit deals +20 damage, while an Extra-Critical Hit deals +30 damage.
```
rosebrallier: I'd use a semi-colon instead of ", while" but I'm not sure whether that's just b/c I love semi-colons.

rosebrallier (approved):
I LOVE THE TABLE SO MUCH
I have some wording suggestions and one big grammar fix.
[edit: sp]

01_02_Combat_Rules.txt
Turtlelord26: I'm going to have to make a rare departure from my usual support of semicolons to question whether using one here would disrupt the rules' flow.

rosebrallier: I think I just don't like "while" here b/c of the time connotation?

Turtlelord26: "and"?

rosebrallier: Oh. Yeah. That.

Issue #529, opened 2018-06-14
Mendar is more op than Kensington at gaining followers...


Wildlovelies:
Mendar's colorful! Ability has no duration and no save after initial... so If I woo a crowd, do they follow me forever?

dfuchs2:
Looks that way. Maybe they should roll a will save every turn?

Turtlelord26:
Every turn is a lot of dice. How about every few minutes, plus a explicit break on entering combat?

bleehu:
Seems to me like it's an instant. So you'd have to roll every time you want to either calm or rile up a crowd you'd need to roll. But having a crowd riled or calmed is not the same as wooing a crowd into followers.

bleehu:
https://github.com/bleehu/Compound_X/blob/dev/CharacterCreation/01_Species.txt#L292
for clarity

Turtlelord26:
Ah, so it's intended to be more like a diplomacy check than a dominate person spell. That's how I thought at first but Wild's question seemed plausible...maybe we need a small quantity of guardrail text?

bleehu:
I think so; it wasn't a wild leap for Wild to make. at the very least, we should say what happens specifically if a will save is failed.

bleehu:
@dfuchs2 can you tackle this?

PR #657, opened 2018-08-16
Clarify Mendar Ability 'Colorful!'

dfuchs2:
Mendar's Colorful! Ability could be misinterpreted to allow Opaleites to turn people into followers

https://github.com/bleehu/Compound_X/issues/529


bleehu (approved):
Great work! Shipit, and we'll close out that issue! One more gripe down!

PR #658, opened 2018-08-16
Update Constructor Drone First Trait

dfuchs2:
Reformatted such that the attribute modifiers were in the sub-species, rather than in the trait. 

Changed +20 to advantage, as it simplifies gameplay somewhat.


bleehu (approved):
Looks good! Shipit!

PR #659, opened 2018-08-16
Update Constructor Drone Second Trait

dfuchs2:
Addressed the problems raised by giving a flat bonus to Jury-Rigging by changing it to advantage.

Removed the demolition part of the trait as it didn't pertain to the second part. Even without it, the trait is pretty powerful.


bleehu:
In future, updates to your branch following requests for changes can be done on the same branch.

bleehu (approved):
shipit! Good work!

Issue #605, opened 2018-07-25
Add Combat Rules for conncurant d10 rolls


bleehu:
As part of our design discussion on Tuesday, we've elected to keep rolling d10s in combat, but we need some rules for rolling a bunch of them at once for automatic fire. 
These rules should cover:
* rolling a jam
* rolling a crit
* rolling more than one of each or both

PR #660, opened 2018-08-16
Add Automatic Fire Concurrent d10 rules

dfuchs2:
Added rules for concurrent d10 automatic-fire.

### DIFF
--- a/01_02_Combat_Rules.txt
+++ b/01_02_Combat_Rules.txt
@@ -609,6 +609,15 @@ miss chance. Lets say your normal ranged miss chance is 2, and the enemy is in
 full cover, your miss chance is increased by 4, for a final DC of 6. So a roll 
 of 7-10 would be successful.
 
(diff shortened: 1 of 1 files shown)


dfuchs2:
Also, @bleehu in https://github.com/bleehu/Compound_X/issues/605 you suggested that these rules cover rolling a jam, crit, and more than one of each or both. I noticed today that automatic fire does not have the ability to crit https://github.com/bleehu/Compound_X/blob/06533a683c7ef51931e6353a90ad7c564e9aa530/01_02_Combat_Rules.txt#L465 Is that still the correct rule?

Reveraine:
That rule is still correct, yes.

bleehu:
Linking Issue #605

bleehu (approved):
elegant, simple, shipit!

Turtlelord26 (approved):
👍

Issue #655, opened 2018-08-14
Hunter dog needs awareness stat


bleehu:
An easy fix, but one I should implement nonetheless.

PR #661, opened 2018-08-17
added awareness stat for hunter dog

bleehu:
Should handle Issue #655. I just plumb forgot that we'd added that stat.


dfuchs2 (approved):
👍

PR #662, opened 2018-08-17
Merge Constructor Drone to Races_Push

dfuchs2:
Merging first and second traits as well as moving attributes to the correct position.


bleehu (approved):
shipit! I'd like another available trait, because I think having 2 for a pick-2 system is a little constrained, but you can still pick the species-wide ones and that would be scope creep. Good work, DFuchs!

PR #665, opened 2018-08-21
Added Pho-Ox

bleehu:
Added the poison Pho-Ox, a Nettle Agent. This is based on the real-life
organic chemical weapon, Phosgene oxime which, fun fact as opposed to VX
gas is abrieviated as CX! This chemical is actually corosive to metal,
but in this ability I've radically exagerated how much damage pouring it
into a robot would do with the conciet that this is an unspecified
 derivative of Phosgene Oxime, not the chemical itself.

This poison is to be unlocked at level 3 as an alternative to Harmasist
Spike.
	modified:   Doctor Procedures.txt

Should handle Issue #561

### DIFF
--- a/CharacterCreation/Class-Specific Documentation/Doctor Procedures.txt
+++ b/CharacterCreation/Class-Specific Documentation/Doctor Procedures.txt
@@ -101,6 +101,81 @@ payload of your choice.
     At level 10, has AP level 6 instead. At level 15, has AP level 8 and deals 
 65 damage. At level 20, ignores physical armor and deals 80 damage.
 
+== Poison: Pho-Ox ==
+
+Adrenaline Cost: 30 Points
+Contagion Cost: 1 Contagion
+Duration: 10 Rounds
+Range: Payload (use in combination with other Proceeedures)
+
+    This dichloroformaldoxime derivative is known as a "Nettle Agent" which 
+means that only extremely high doses will kill. Some might call that mercy, 
+while those it has been used on know better. Organics exposed to either the 
+liquid (which is a dull amber color with an oily texture) or the vapor almost
+immediately break out in hives, itch all over, serious eye irritation and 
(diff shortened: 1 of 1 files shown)


bleehu:
Looking at very-low level Doctor, I'd like to either give them their first Procedure at level 1, or let them start out with extra Med Kits. But as it's written, Doctors get Procedures as they level up.

dfuchs2 (approved):
This is awesome! I like the idea of a poison that affects drones as well as organics.

CharacterCreation/Class-Specific Documentation/Doctor Procedures.txt
```diff
@@ -101,6 +101,81 @@ payload of your choice.
...
+make a Shock save of DC 80 at the begining of their turn to reduce the dose 
+level by one. Every time the dose level is increased beyond 6, the target makes
+a Shock Save of DC 110 instead of 70. If they succeed, they maintain their dose
+level. If they fail, they fall unconsious (but not bleeding out) and will not
+wake on their own for 8 + 1d10 hours. Non-organic targets affected by Pho-Ox do
+not take any penalties to accuracy or move speed, but instead take only damage
+at the start of their turn. Targets that are wearing a gas mask, but have any 
+exposed biology or unarmored cybernetics are affected by this even if a 
+deployment method says they wouldn't be. Pho-Ox can be rapidly cured in the 
+field by most combat medics.
```
Turtlelord26, line 136: nb to dev, will block to RC: I would like to see a Medicine (Int) DC to execute that cure.

Turtlelord26 (approved):
V shiny, although we're never getting to the high dose levels on your average PC-sized health pool robot unless they're getting a lot of healing. Save and duration might be a bit high.

Numbers subject to testing, concept excellent 👍.

Issue #264, opened 2017-12-29
Need New Doctor Attack Procedure To Replace Pasteurize


Turtlelord26:
Pasteurize exists as a node in the tree, just above Harmacist Spike, but has no entry in the descriptions document.

bleehu:
Needs to be replaced by a new attack power; pasteurize as it was written is out of aesthetic.

bleehu:
Need to finish merging PR #668 so that we don't overwrite the binary image file for the doctor procedure tree; binary files cannot be combined in git.

bleehu:
Closing as done.

PR #669, opened 2018-09-07
Mh/issue 264/replace pasteurize

bleehu:
Issue #264  Added Neural Afib ability to replace pasteurize. 
I'm a little concerned that giving the Doctor a single target Disrupt ability detracts a little from Hacker. 
Disrupted deffinition is here: https://compoundx.org/docs/conditions#7
The idea of a single target-debuff in terms of Doctor Design is here: https://github.com/bleehu/CX_Design_Guide/blob/master/characters/how_to_classes.md#to-modify-doctor


Turtlelord26:
Alarm: you based off of play.

Turtlelord26 (approved):
Mildly concerned at the duration, which seems long to me. However, seems like a good.

sp: Afibrillator.

PR #680, opened 2018-09-15
Changed style of Doctor Class Feats

bleehu:
and added flavor text to the Doctor class

	modified:   02_Classes.txt

### DIFF
--- a/CharacterCreation/02_Classes.txt
+++ b/CharacterCreation/02_Classes.txt
@@ -144,10 +144,11 @@ Skill Proficiencies:
 
 =====   DOCTOR   =====
 
-    You know the secrets of life and death. The very continued existance of 
-those around you depend on you striking the right balance between heal and harm. 
-One errant twitch of your hands could mean the difference between saving the day 
-or losing a friend. This is what you've been training for for years. 
+    You know the secrets of life and death. The very continued existence of 
+those around you depends on whether you strike the right balance between heal 
+and harm. One errant twitch of your hands could mean the difference between 
+saving the day or losing a friend. This is what you've been training for for 
+years. 
 
 Class Primary Attribute: Dexterity
 Starting Wealth: $2670 + 100*Luck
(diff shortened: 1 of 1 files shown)


Turtlelord26 (commented):
What do you think about copying the prerequisite format from the rest of the feats?

Aside from that, my only comments are
* This looks good
* 'Errant twitch' in lines 149 and 190 is repetitive.

bleehu:
My concern with copying the prereq format is that we have a lot of feats with only one prereq, and the parser needs at least 2 to realize that we're making a list.

CharacterCreation/02_Classes.txt
```diff
@@ -143,8 +143,11 @@ Skill Proficiencies:
 
 
 =====   DOCTOR   =====
-You know how or why something works [most likely humans / humanoid bodies), 
-and put that knowledge to good use by supporting your team.
+
+    You know the secrets of life and death. The very continued existance of
```
rosebrallier: sp:"existence"

CharacterCreation/02_Classes.txt
```diff
@@ -143,8 +143,11 @@ Skill Proficiencies:
 
 
 =====   DOCTOR   =====
-You know how or why something works [most likely humans / humanoid bodies), 
-and put that knowledge to good use by supporting your team.
+
+    You know the secrets of life and death. The very continued existance of 
+those around you depend on you striking the right balance between heal and harm.
```
rosebrallier: gr:"depends"
Also, I'd go with "on whether you strike the right balance between heal and harm."

CharacterCreation/02_Classes.txt
```diff
@@ -143,8 +143,11 @@ Skill Proficiencies:
 
 
 =====   DOCTOR   =====
-You know how or why something works [most likely humans / humanoid bodies), 
-and put that knowledge to good use by supporting your team.
+
+    You know the secrets of life and death. The very continued existance of 
+those around you depend on you striking the right balance between heal and harm. 
+One errant twitch of your hands could mean the difference between saving the day
```
rosebrallier: "one errant twitch" is a vivid enough phrase that I'd only use it once in a short passage like this. m/b choose whether to use it here or in the Adrenaline Junkie flavor text.

rosebrallier (approved):
Shipit! The new prose really captures the central conflict of the class (to f- people up or un-f mah frands)

rosebrallier:
Oh! Just to make sure, the 2nd person is an intentional deviation from the norm, yes? (It works really well imo)

PR #688, opened 2018-09-16
rewrote Marksman Class Feats

bleehu:
the old ones were out of aesthetic.
The new Marksman class abilties lean into their role at eleminating
key targets such as enemy medics, hackers, etc.
	modified:   02_Classes.txt


Turtlelord26 (approved):
Glorious! Such Aesthetic!

PR #685, opened 2018-09-16
rewrote Hacker Class description and feats

bleehu:
added `Characters/Class-Specific/Hacker Exploits.txt` This is not meant
to be a long-term rewrite of the class, but instead a formalizing of
the existing hand-wavy makeshift system in place already.
This is why we are missing numbers for disrupting weapons, cameras or
enemy security drones for instance.

	modified:   02_Classes.txt
	new file:   Class-Specific Documentation/Hacker Exploits.txt

PR #697, opened 2018-09-16
Pwn pronunciation guide

Turtlelord26:
Unresolved from #685


bleehu:
I think the hard "o" symbol might break the encoding?

Turtlelord26:
I know it breaks ANSI but is fine in Unicode. Which are we?

bleehu:
I guess we are unicode. Let's give it a shot! 
http://flask.pocoo.org/docs/1.0/unicode/

bleehu (approved):
shipit

Issue #599, opened 2018-07-19
Gripe: Flamethrower Aoe


rosebrallier:
Flamethrowers should have an Aoe option.

bleehu:
@dfuchs2 Do you think you can take this one? Or are you swamped?

Turtlelord26:
I rather like the lv 1 flamethrower as a single-target weapon that shoots smaller globs of fast-burn-up gel into faces. That said, we're also missing a higher-level flamethrower. Can we solve two problems at once here?

dfuchs2:
Doesn't look like anybody has taken this one yet. I'll do it next sprint. I very much like the idea of a higher-level flamethrower that acts differently than the one we already have.

PR #699, opened 2018-09-25
Add AoE Flamethrower

dfuchs2:
Added a second flamethrower with AoE

### DIFF
--- a/Items/04_06_MachineGuns_HeavyWeps_&_RocketLaunchers.txt
+++ b/Items/04_06_MachineGuns_HeavyWeps_&_RocketLaunchers.txt
@@ -93,32 +93,6 @@ Movement Speed Penalty: -2m
 * 2 out to 15m
 * 4 out to 45m
 * 5 out to 80m
-
-==Frontier Flame Thrower FLMR-B $3880 ==
-
-Damage: 80 Special
-Mag Size:  8 units of 6 liters, Naphitic Gel. $320 per magazine
-Reload DC:  50
-Fire Modes:  Semi
-Armor Piercing Level: n/a
-Reflex Modifier: -20
-Suppression Level: Medium
-Movement Speed Penalty: -2m
-
-	This flame thrower fires a wide spread of flaming gel, hitting the target 
-space as well as each adjacent space visible to the wielder.
-	Catches flammable objects on fire. Not affected by armor, unless target is 
-100% sealed. Any targets that took damage apply 1 x Burning @ DC70% to remove. Apply 1 x
-Fear on hit. No optics or barrel attachments.
-
-	Ranged Miss Chance
-
-* 2 out to 10m
-* 5 out to 15m
-* 6 out to 20m
-
-    The Frontier Flame Thrower FLMR-B has large magazines that take up 2 magazine 
-slots.
  
 ==Matthews Model 240-B  $4450 ==
 
@@ -319,6 +293,37 @@ listed value when used in this gun. 2 action reload time.
 * 3 out to 15m
 * 4 out to 40m
 * 5 out to 100m
+
+==Frontier Flame Thrower FLMR-B $3880 ==
+
+Damage: 80 Fire
+Mag Size:  8 units of 6 liters, Naphitic Gel. $320 per magazine
+Reload DC:  50
+Fire Modes:  Semi
+Armor Piercing Level: n/a
+Reflex Modifier: -20
+Suppression Level: Medium
+Movement Speed Penalty: -2m
+
+	This flame thrower fires a wide spread of flaming gel, hitting the target 
+space as well as each adjacent space visible to the wielder. The flaming gel 
+continues to burn for one round of turns.
+	Catches flammable objects on fire. Not affected by armor, unless target is 
+100% sealed. Any targets that took damage apply 1 x Burning @ DC70% to remove. 
+Apply 1 x Fear on hit. No optics or barrel attachments.
+	When the flame hits its target space, anybody who takes a step into a 
+space that has flame on it or begins their turn in one must make a DC60 DEX roll 
+to avoid taking damage from lingering flame on the ground. Failing this roll 
+deals 30 damage and 1 x Burning @ DC70% to remove.
+
+	Ranged Miss Chance
+
+* 2 out to 10m
+* 5 out to 15m
+* 6 out to 20m
+
+    The Frontier Flame Thrower FLMR-B has large magazines that take up 2 magazine 
+slots.
 
 ==Anvil Arms XM-500 Minigun $4477 ==
 
--- a/Items/04_06_MachineGuns_HeavyWeps_&_RocketLaunchers.txt
+++ b/Items/04_06_MachineGuns_HeavyWeps_&_RocketLaunchers.txt
@@ -94,6 +94,32 @@ Movement Speed Penalty: -2m
 * 4 out to 45m
 * 5 out to 80m
 
+==Frontier Flame Thrower FLMR-B $3880 ==
+
+Damage: 80 Special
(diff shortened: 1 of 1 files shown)


dfuchs2:
As griped in https://github.com/bleehu/Compound_X/issues/599

Items/04_06_MachineGuns_HeavyWeps_&_RocketLaunchers.txt
```diff
@@ -94,6 +94,32 @@ Movement Speed Penalty: -2m
...
+Mag Size:  8 units of 6 liters, Naphitic Gel. $320 per magazine
+Reload DC:  50
+Fire Modes:  Semi
+Armor Piercing Level: n/a
+Reflex Modifier: -20
+Suppression Level: Medium
+Movement Speed Penalty: -2m
+
+	This flame thrower fires a wide spread of flaming gel, hitting the target 
+space as well as each adjacent space visible to the wielder.
```
bleehu: For clarity, can you catch the square on fire, but miss the target if you don't roll accuracy high enough? What happens to a character that enters, leaves, moves through or ends their turn in a square that is burning? How long does the burning last?

Items/04_06_MachineGuns_HeavyWeps_&_RocketLaunchers.txt
```diff
@@ -94,6 +94,32 @@ Movement Speed Penalty: -2m
 * 4 out to 45m
 * 5 out to 80m
 
+==Frontier Flame Thrower FLMR-B $3880 ==
```
bleehu: That's a steep price hike over the level 1 flamethrower. So long as it's worth what you're paying for that's fine.

Items/04_06_MachineGuns_HeavyWeps_&_RocketLaunchers.txt
```diff
@@ -94,6 +94,32 @@ Movement Speed Penalty: -2m
 * 4 out to 45m
 * 5 out to 80m
 
+==Frontier Flame Thrower FLMR-B $3880 ==
+
+Damage: 80 Special
```
bleehu: 20 less damage than the level 1 flame thrower? I suppose there's the potential to deal it to multiple targets, so maybe that makes sense. Can we change it to deal fire damage? https://github.com/bleehu/Compound_X/blob/play/01_04_Damage_Types_and_Armor_Types.txt#L51

Items/04_06_MachineGuns_HeavyWeps_&_RocketLaunchers.txt
```diff
@@ -94,6 +94,32 @@ Movement Speed Penalty: -2m
 * 4 out to 45m
 * 5 out to 80m
 
+==Frontier Flame Thrower FLMR-B $3880 ==
+
+Damage: 80 Special
+Mag Size:  8 units of 6 liters, Naphitic Gel. $320 per magazine
```
bleehu: I like the relatively low ammo capacity. This thing hits like a truck. Slowing the user down to reload often I think creates a beautiful cadence for combat.

bleehu (changes requested):
Blocker: flamethrower added in the machine gun section.

Items/04_06_MachineGuns_HeavyWeps_&_RocketLaunchers.txt
dfuchs2: I hadn't considered that. I imagine the flame would burn on the ground for one additional turn. I was imagining it as a faster-burning napalm. I'm thinking of making it such that the player rolls dex for each burning space they step in, including the one they were in when the flame thrower hit. Anybody who ends their turn in the fire would roll dex again one more time before the fire diminishes. The burning damage from the fire on the ground would be in the range of 30-40.

dfuchs2:
I can't believe that I put it in the wrong section. That was definitely not intentional.

Items/04_06_MachineGuns_HeavyWeps_&_RocketLaunchers.txt
dfuchs2: I see the original flame thrower is special damage. Should that also be fire damage?

dfuchs2:
As a side note, I was wondering about using flamethrowers in a vacuum.  In reality, flamethrowers were commonly used to deoxegenate bunkers and caves. Since there is no oxygen in space, do our flamethrowers work there? Do they have their own oxygen supplies?

bleehu:
That is a very good question! There's a couple of different directions you could take this. 
1) fire needs oxygen. No oxygen, no flamethrower. Tactical puzzle time!
2) flamethrower propellant is made of stuff that has its own oxygen chemically suspended in the gel like napalm does. It burns in space, under water, and that's pretty much all there is to it. Keeps the mechanics nice and simple.
3) fire burns more quickly with more oxygen, maybe using the flamethrower in an airless environment only reduces the flame counters it gives and maybe it reduces the damage?

Items/04_06_MachineGuns_HeavyWeps_&_RocketLaunchers.txt
```diff
@@ -320,6 +294,37 @@ listed value when used in this gun. 2 action reload time.
...
+Reload DC:  50
+Fire Modes:  Semi
+Armor Piercing Level: n/a
+Reflex Modifier: -20
+Suppression Level: Medium
+Movement Speed Penalty: -2m
+
+	This flame thrower fires a wide spread of flaming gel, hitting the target 
+space as well as each adjacent space visible to the wielder. The flaming gel 
+continues to burn for one round of turns.
```
bleehu, line 310: Very specific! I like it!

Items/04_06_MachineGuns_HeavyWeps_&_RocketLaunchers.txt
```diff
@@ -320,6 +294,37 @@ listed value when used in this gun. 2 action reload time.
...
+	Catches flammable objects on fire. Not affected by armor, unless target is 
+100% sealed. Any targets that took damage apply 1 x Burning @ DC70% to remove. 
+Apply 1 x Fear on hit. No optics or barrel attachments.
+	When the flame hits its target space, anybody who takes a step into a 
+space that has flame on it or begins their turn in one must make a DC60 DEX roll 
+to avoid taking damage from lingering flame on the ground. Failing this roll 
+deals 30 damage and 1 x Burning @ DC70% to remove.
```
bleehu, line 317: simple succinct and I don't think it can be misinterpreted!

bleehu (approved):
It seems aesthetically good to me, I like the mechanics. Shipit!

Items/04_06_MachineGuns_HeavyWeps_&_RocketLaunchers.txt
Turtlelord26: We can punch the stats into 1source's spreadsheet and see what it tells us 🐢

Turtlelord26:
Notes to self: rules for aoe and relevant damage types in nonstandard gravitatational/atmospheric conditions. We shouldn't worry about those in individual weapon descriptions; they should be more general.

Turtlelord26:
I kinda like the idea that blee's number one is a default but there's a specialty ammo to make number two happen.

dfuchs2:
Perhaps we should stick with number one, and then add number two as a fuel/oxidizer mix? We could increase price by 1.5x and reduce ammo per magazine by 1/3? (Due to the added mass of our oxygen)

Turtlelord26:
New issue outside this immediate scope

dfuchs2:
True. We can add that in as a new ammo type, but this weapon is probably done as is.

bleehu:
Than hit that big green merge button Shrub! You earned it! :)

PR #701, opened 2018-09-28
Add vacuum flame thrower fuel

dfuchs2:
Not sure whether to make it level 1 or level 10, as I don't know the criteria used when the other ammo types were made.

### DIFF
--- a/Items/04_00_WeaponAttachments.txt
+++ b/Items/04_00_WeaponAttachments.txt
@@ -694,6 +694,17 @@ using Vorpal rounds, your gun deals 80 + 20 (for Critical) + 8 (Vorpal damage) =
 If your roll an Extra Critical Hit, you would deal 80 +  30(for Extra Critical) 
 + 24 (Vorpal Damage) = 134 damage.
 
+== Naphoxic Gel ($1.5 x mag cost) == 
+    
+    Naphitic gel with small oxygen capsules dispersed inside. Standard flame thrower 
+gel requires atmospheric oxygen to burn, and as such cannot be used in a vacuum. 
+Naphoxic gel can be burnt in any environment, including space and underwater. 
+While in atmospheres with oxygen, it burns hotter than normal gel. The added oxygen
+in the tank reduces its overall fuel capacity.
+    
+    Ammo Capacity: -20% (rounded down)
+    Base Damage: +10% (in atmosphere)
+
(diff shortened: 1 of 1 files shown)


Items/04_00_WeaponAttachments.txt
```diff
@@ -694,6 +694,17 @@ using Vorpal rounds, your gun deals 80 + 20 (for Critical) + 8 (Vorpal damage) =
 If your roll an Extra Critical Hit, you would deal 80 +  30(for Extra Critical) 
 + 24 (Vorpal Damage) = 134 damage.
 
+== Naphoxic Gel ($1.5 x mag cost) == 
+    
+    Naphitic gel with small oxygen capsules dispersed inside. Standard flame thrower 
+gel requires atmospheric oxygen to burn, and as such cannot be used in a vacuum. 
+Naphoxic gel can be burnt in any environment, including space and underwater. 
+While in atmospheres with oxygen, it burns hotter than normal gel. The added oxygen
+in the tank reduces its overall fuel capacity.
```
bleehu, line 703: "reduces fuel capacity" This is brilliant! It's so simple, so obvious, but I'd have never thought of it! And It plays right into one of the few drawbacks of flamethrowers as they are already! Genius!

bleehu (approved):
shipit!

Turtlelord26:
My thoughts and concerns:

Level 1 is a good place to put the ammo, as it doesn't restrict DM options of what kinds of dungeons to give their characters. We are a space sci-fi game, so we should allow play in space at any level.

I don't like the ammo capacity reduction on the ammo. Going back to the notional lv dungeon *in space!*, the drawback combined with the extra cost feels like a tax. No one likes taxes. I would instead suggest that we reduce the cost multiplier (though keep it above one) and instead of a drawback or in addition to it, find some (more universal) bonus we could give the ammo. This bonus would not by itself be worth the extra cost, because there should be some cost to using fire in a null-oxygen environment, but it would still feel like you were getting something for your money instead of being taxed.

Line 700: "cannot be used in a vacuum". would prefer "cannot be used in a low- or zero-oxygen environment.".

Finally, we should have a development guideline that states that when you ask for reviews from multiple people, you shouldn't merge until you either get all requested reviews or have waited 24-48 hours (in addition to the requirements already in place). @bleehu

PR #709, opened 2018-11-02
deprecated the unbalanced cyborg species

bleehu:
modified:   CharacterCreation/01_Species.txt

I've done some clever hotfix git magic that will make this change easily ahead of the regular push from dev->RC so that Kaila doesn't try to make a cyborg during our play/test session this weekend. The hotfix is using the same commit that I used in dev, so when the RC version bump does come, it should seemlessly blend together as if we'd shipped it normally.


Turtlelord26 (approved):
Teach me your glorious lizardry, master lizard Bleehu.

Reveraine (approved):
THROW IT IN THE FIRE

bleehu:
Throw it in the fire

Issue #711, opened 2018-11-04
Feature Request: Balanced Stable of Melee Weapons of Various Sizes


Turtlelord26:
Berserkers will thank us.

But more generally, we need more melee weapons to fill out all of the size categories at each level bucket.

PR #712, opened 2018-11-04
Adding the Whackbush Arms Spear, a lv 1 2-h weapon

Turtlelord26:
Price is pretty arbitrary.
Damage numbers come out to 84 with 7Str and 7Dex and 120 with 9Str and 7Dex.

Drawback comes from concerns about the lv 1 damage potential, and in particular oneshotting unarmored enemies. It may be removed later to be replaced with a more general drawback for two-handers.

solves the gripe portion of #711


bleehu (approved):
Let's give it a try!

Issue #670, opened 2018-09-07
Visibility Rules


Reveraine:
We don't have written rules on how far a normal individual can see in various lighting conditions.

Issue #509, opened 2018-05-14
Gripe Feature Proposal: Light-amplifying Goggles


Turtlelord26:
Would allow normal vision in low-light conditions. Specifically does not use Infrared and does not "enhance" imagery, instead uses light-gathering and amplifying technology to make the view appear brighter than it is.

bleehu:
What effect would this have on watching a flashbang go off?

Turtlelord26:
[Wince]

PR #713, opened 2018-11-07
Create Light-Amplification Goggles

dfuchs2:
A cheaper variant of night vision goggles that does not use IR. Very weak to flashbangs.

### DIFF
--- a/Items/07_Items.txt
+++ b/Items/07_Items.txt
@@ -65,6 +65,12 @@ chip that helps you find your dog if it runs away. The harness itself is
 lightweight and a little bit stretchy to prevent irritating the wearer. Also
 comes with a 3m leash. 
 
+Light-Amplification Goggles ($150):
+    Light-amplified visibility in low light conditions. All light is amplified, 
+unlike previous monochrome models. If the user is blinded by a flashbang 
+grenade, they gain an extra turn of blindness and cannot make the blindness 
+awareness save for their next turn. Resolution clouds after 80m.
+
 Night Vision Goggles ($300):
     IR visibility for low light conditions. Resolution clouds after 100m.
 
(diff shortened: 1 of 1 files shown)


bleehu (approved):
I like it! Shipit!

Turtlelord26 (changes requested):
I would like the item to explicitly state that the user gains the ability to see in low-light conditions as if they were normally lit (if they do not already have that ability). Otherwise, shipit the shiny thing!

*The reason being that when our item has a mechanical effect beyond RP usage, we should use mechanical text to make that clear.

Turtlelord26:
Better, and possibly the root problem here is that we don't have mechanically well-defined light levels, but your new text is almost indistinct from night vision goggles.

dfuchs2:
Good point. Should we add some extra documentation to the night vision goggles to make them distinct? For example, IR should be able to see through smoke pretty clearly.

Turtlelord26:
My original vision was not to provide darkvision with the light-amplifying goggles. I was thinking of a set of relations like below:
pitch black -> still pitch black
night -> low-light
low-light -> day
day -> still day

As distinct from NVG, which do 
all conditions -> monochrome day (and occluded by different types of obstacle)

In this world, neither item is strictly better than the other, as they have different use cases.

You're absolutely right that NVG needs more text too, but that's a separate problem / new issue.

Since you're the one doing the work, you're free to make a case for the LAG to have a different functionality than the original intention, especially since I didn't spell it out very well in #509. I hope it helps to know where I'm coming from with the reviews though.

dfuchs2:
In that case, I think these goggles might be blocked by https://github.com/bleehu/Compound_X/issues/670 . I'll work on that one next sprint. To my knowledge, we don't have good definitions of darkness levels anywhere yet.

Issue #716, opened 2018-11-12
Gripe: Melee Crit Fail Checks


Reveraine:
Extend melee crit failure rules for improvised melee firearms to include maneuver checks.

PR #717, opened 2018-11-12
Add Critical rules to maneuver checks.

Turtlelord26:
#716 

It's unusual for a skill check, but in line with the other combat rolls. Also I kinda want someone to accidentally discharge a pistol while grappling and trigger the shooting-into-a-melee rule *on themselves*.

Also moved the improvised weapon crit fail rule to the critical section, from the improvised section, since that section seems more likely to be visited during play.


bleehu (approved):
seems sane. Shipit!

PR #725, opened 2018-12-08
Medicalize Adrenaline Rush

Turtlelord26:
Some functional change to adrenaline rush during the conversion.
Now only gives either +4 of +8 instead of anything between +1 and +10.
Also removed the take-the-feat-twice-to-halve-the-cost, but I could be persuaded to restore that.


bleehu (approved):
shipit!

PR #727, opened 2018-12-08
Medicalize Bunker

Turtlelord26:
Bunker only works with nanite space magic, and I am considering it as out-of-aesthetic.


bleehu (approved):
So... there's a story here. 
"Bunkering" is a paintball turn which loosely translates to running up to a bunker and blind-firing at near-point-blank at the enemy who's on the other side of the same cover as you are. This apparently got horribly lost in translation when I suggested the feat. I'd like to add an issue to put a non-magic bunkering active feat in that bears little resemblance to this wizardry.

bleehu:
Here's a video with a couple of bunker maneuvers in it. A couple of cover slides as well! https://www.youtube.com/watch?v=nAFNAL1zVl0

Turtlelord26:
Do the thing!

PR #729, opened 2018-12-08
Medicalize Cover Slide

Turtlelord26:
Some functional changes here. 
* Now does not require that you end in a space with cover, to make tracking easier for the DM.
* Now consumes two actions instead of one, to avoid slapping a huge cooldown on it.
* Now gives an evasion bonus to compensate for the above. Useful for dodging readied shots and light suppression when moving between covers.

General note on Active Feats with a cooldown of None: I gave this to a few of them, generally because I'm not concerned about multiple uses per round (in the case of cover slide, it would be really hard to do so at all. The intention is that the Cooldown: None feats still require you to have a unit of medi nanites available and off cooldown., it just doesn't then send them into any remission.

### DIFF
--- a/CharacterCreation/03_Feats.txt
+++ b/CharacterCreation/03_Feats.txt
@@ -934,14 +934,19 @@ the action after this feat ends. If this malus would not run out until after
 this turn, its effect wraps, around to the beginning of your next turn. Cannot 
 be used in concert with Run 'N' Gun.
 
-== Cover Slide [15 Nanites] ==
+== Cover Slide ==
+
 Prerequisites:
 
 *  DEX 4
 
-	Move up to your speed and end next to a barrier behind which you can take
-cover. You hunker or go prone behind that cover as a free action. Cannot 
-be used in concert with Run 'N' Gun.
+Cooldown: None
+Duration: 2 Actions
+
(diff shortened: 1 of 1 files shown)


bleehu (changes requested):
Totally agree on a cooldown of none; But I would really really like to see the cover part of the cover slide retained, as well as the one-action limit; Since moving your speed is one action, and going prone is one action, combing those in a two-action feat becomes of negligible worth.

Turtlelord26:
Moving your speed takes two actions. Will discuss other points in person.

Turtlelord26:
Now only uses 1 Action, but gained a cooldown to compensate. Do you still want to restrict it to ending in cover? I'm a little worried about players and DMs disagreeing on what constitutes cover.

bleehu:
They might, but we define cover pretty well in the combat rules already, https://www.compoundx.org/docs/combat#152. I would like it to continue to require cover, but after our discussion last night, it sounds like I'm the only one. Do what you feel is right.

bleehu (approved):
:shipit:

Turtlelord26:
Actually, think your demonstration of the momentum stop convinced me. What do you think of this wording?

PR #730, opened 2018-12-08
Medicalize Flash Dance

Turtlelord26:
Spin to win?

Hoo-kay, I tried to fix this one while I was in here. Here's what I changed:
* Now has a duration between 2 and 6 actions, instead of 1 action. Cooldown scales to duration.
* Range of flash dance reduced from  '"a 5m space" that you could be adjacent to' to a 2m radius centered on the user
* In compensation for the range nerf, I've allowed the use of several combat maneuvers in addition to attacks. I specifically left out the ones that require the user to move, and also grapple because that would end the flash dance.
* added explicit flash dance termination due to stun.

### DIFF
--- a/CharacterCreation/03_Feats.txt
+++ b/CharacterCreation/03_Feats.txt
@@ -943,16 +943,21 @@ Prerequisites:
 cover. You hunker or go prone behind that cover as a free action. Cannot 
 be used in concert with Run 'N' Gun.
 
-== Flash Dance [10*X Nanites] ==
+== Flash Dance ==
+
 Prerequisites:
 
 *  Level 3
 
-	For 2 Actions, the user can make Lvl/2 (round up) number of attacks, up to a 
-max of 6, with their melee weapon(s). The attacks must take place within a 5m 
-space, rolling for each strike, where X is the number of attacks. All bonuses 
-that would normally apply to a standard attack still apply. (Note: Flash Dance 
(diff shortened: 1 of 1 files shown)


CharacterCreation/03_Feats.txt
```diff
@@ -943,16 +943,21 @@ Prerequisites:
...
-can trigger Bulltrue and Preemptive Strike)
+Cooldown: 2 * X + 1 Rounds
+Duration: 2 * X Actions, up to X = 3
+
+	For the duration of Flash Dance, every 2 Actions you may instead make a 
+number of Melee Attacks, Trip, Disarm, or Kick attempts equal to your level divided 
+by 2 (rounding up), to a maximum of 6. These actions have a range of 2m. Roll 
+separately for each act. All bonuses that would normally apply to a standard attack 
+still apply. (Note: Flash Dance can trigger Bulltrue and Preemptive Strike). If you 
+are stunned or rendered unconscious, Flash Dance ends immediately.
```
bleehu, line 960: I really like the new explicit termination cases.

bleehu (approved):
I'm a little hesitant about the complexity creep here, but Rev says that the equations aren't so complicated that the average player can't follow them, so I'm not gonna worry about it. You're a braver dev than I trying to preserve flashdance, and I think the game is better for your work!

PR #731, opened 2018-12-08
Medicalize Gun Kata

Turtlelord26:
Made a couple changes.
* Now consumes Reactions instead of being an Action. I did this to mirror Kata.
* Now limits targets to a 45 degree cone, to compensate for the improved action economy
* Added explicit termination on stun
* Dropped the CQB function, because it doesn't seem like it does very much. If you need your gun to be CQB and it isn't already, you aren't going to get the first shot off to trigger Gun Kata in the first place, right?


bleehu (approved):
:shipit:

PR #732, opened 2018-12-08
Medicalize Kata

Turtlelord26:
Changes:
* Now consumes Reactions instead of granting implied free action attacks.
* Now supports some combat maneuvers in addition to attacks, mirroring new Flash Dance.


bleehu (approved):
Love it! Shipit! :shipit:

PR #733, opened 2018-12-08
Medicalize Kip Up

Turtlelord26:
An interesting one, to be sure. Had to preserve the unique cooldown structure, because it's great.

Changelog:
* Now cannot kip up in microgravity. Because you don't need a feat to do that.
* Now also cannot kip up in power armor. It doesn't exist yet, but one day... (I'm, sure we'll make a module that lets the power armor kip up, but we can call it Powered Kip Up or something)
* The 3 round cooldown is harsher than I would normally apply, but I think that it will help with tracking. This feat has more than its fair share of tracking complexity already.


bleehu (approved):
The two uses, 3-round cooldown seems a little needlessly complex to me, but it's by no means a blocker. Shipit!

PR #734, opened 2018-12-08
Medicalize Multi-Target

Turtlelord26:
I can't figure out a flavor for this feat that isn't nanite-magic-out-of-aesthetic or redundant with gun kata. So I'm calling it out-of-aesthetic.


bleehu (approved):
I was trying to fight it, but I think you're right. It's magic nanite BS.

PR #735, opened 2018-12-08
Medicalize Run 'N Gun

Turtlelord26:
Wording changes, mostly ordering.
Ever so slightly significant change in that Running and Gunning for 3 actions is now one use of the feat instead of 3 uses, but that made the cooldown a lot less snarly.


bleehu (approved):
noice! Shipit!

PR #737, opened 2018-12-08
Medicalize Twitch and Blur

Turtlelord26:
Twitch's accuracy penalty now has a duration, so you can't stack up a -5000 Accuracy over the course of a campaign. MIN_INTEGER and all that. Now also allows aim at non-enemies. Other than that, no functional changes.

These two are bundled into a single PR because I think they should be reviewed together both initially and later.


bleehu (approved):
love it! Great call on the logical grouping. Shipit!

Issue #611, opened 2018-07-25
Update Active Feats to new nanite system


bleehu:
as part of our push to update the nanite system, we need to change both the nanite costs of active feats and the description of how the nanites work to help the action.

PR #739, opened 2018-12-12
Turtlelord26 #611 active feat nanite removal

Turtlelord26:
#611 
Behold, the de-naniteification of the active feats! Notes are in the commit descriptions, including functional changes.
Please review with an eye to the overall balance of this new resource economy. Thank!

Be not concerned at the line changes, it's git weirdness. The only changes are to active feats.


bleehu (approved):
I would like to see extensive testing, but I think we're ready for that now :)

Issue #722, opened 2018-11-28
Feature Request: Room Clearing Combat Rules


bleehu:
The "Point man" feat and weapon reflex modifiers both point to a cool rule that TRowl came up with around the start of the game that never got written down. There's supposed to be a stack-up and room-clearing system where the players and enemies roll reflex saves against each other to determine who shoots first. For those of you who have played the popular "Rainbow 6 Siege" you should have some idea of how this is meant to go. 

The ask here is to write a few paragraphs under a heading in the combat rules that explains this system in a way that is easy to learn the first time and easy to skim to remwmber if you haven't played in a while.

Stretch goal: breaching charge item that penalizes the reactions of enemies in the room.

bleehu:
Shipp'd

PR #742, opened 2018-12-28
added breach and clear rules

bleehu:
Added rules for breaching and clearing. In case you were wondering why our weapons had reflex modifiers on them.

Issue #722

### DIFF
--- a/01_02_Combat_Rules.txt
+++ b/01_02_Combat_Rules.txt
@@ -5,16 +5,16 @@ Combat
 
 	Combat starts when either you or an enemy decides they would like to make an 
 attack (or as the GM decrees). At that point, all involved characters roll 
-Initiative to determine the order in which they take turns. Initiative is a d100 roll 
-that is modified only by Soldier class passive abilities, and higher numbers go 
-first. The section "Actions", below, has more detail on turns in combat.
+Initiative to determine the order in which they take turns. Initiative is a d100 
+roll that is modified only by Soldier class passive abilities, and higher numbers 
+go first. The section "Actions", below, has more detail on turns in combat.
 
 ===== Combat Stat Modifiers for Ranged and Melee Accuracy =====
 
 	These bonuses use a table instead of the more general attribute modifier 
-equation, as combat accuracy rolls are played out on a d10 scale. As described below, 
-under normal conditions, ranged combat uses the Perception combat modifier, while 
-melee combat uses Strength, Dexterity, and Fortitude.
+equation, as combat accuracy rolls are played out on a d10 scale. Under normal 
+conditions ranged combat uses the Perception combat modifier as described below. 
+Melee combat uses Strength, Dexterity, and Fortitude.
 
 == Combat Stat Modifier Table ==
 
@@ -45,7 +45,7 @@ melee combat uses Strength, Dexterity, and Fortitude.
 	Drop any remaining decimal (0.5 or -0.5 becomes 0) -in other words- only use 
 integers when you are writing down the final miss chances of your weapon(s). For 
 example, let's say you have a basic SMG with miss chance brackets of 4, 5, and 6. 
-Assuming you have 6 Perception,your PER bonus is 0.5 and thus, your gun's range 
+Assuming you have 6 Perception, your PER bonus is 0.5 and thus, your gun's range 
 miss chance brackets are:
 
 	(4 - 0.5), (5 - 0.5), (6 - 0.5)  =  3.5, 4.5, 5.5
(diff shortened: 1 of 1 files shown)


01_02_Combat_Rules.txt
```diff
@@ -999,3 +999,128 @@ retrieve a dropped weapon or stand back up if they have fallen.
...
+clearing actions to re-draw their weapon. As such, devices are usually used by
+the characters in the 3rd or 4th positions.
+
+    Some rooms are defended with booby traps. In the case that a room or entry
+point has such perils, they take effect second. 
+
+    While devices are being resolved, all attackers and defenders roll a reflex
+save to determine the order of combat during the clearing phase. If the 
+defenders were unaware of the impending breach, then they roll reflex with a
+disadvantage and the attackers roll reflex with advantage.
```
Turtlelord26, line 1091: This seems unnecessarily punishing. I would suggest only defenders-disadvantage for being surprised, and leave attackers-advantage for other circumstances to invoke.

01_02_Combat_Rules.txt
```diff
@@ -999,3 +999,128 @@ retrieve a dropped weapon or stand back up if they have fallen.
...
+===== Clearing =====
+
+    Players take their turn in decending order of breaching roll. In other words,
+The character with the highest net reflex roll goes first, the character with 
+the lowest reflex goes last. Characters on the attacking team (the team that 
+stacked up) must use an Action to move from their position in the stack line
+before they can fire or perform melee attacks. Characters on the attacking team
+can take cover at the entry point that they were stacked up on, but doing so 
+will block the entry point and prevent another character from moving or firing
+through the entry point unless they are firing at that character.
```
Turtlelord26: Suggest phrasing: "prevent ... moving through or firing at targets past that character.

01_02_Combat_Rules.txt
```diff
@@ -999,3 +999,128 @@ retrieve a dropped weapon or stand back up if they have fallen.
...
+
+    The Clearing phase ends when either all of one team in the room can no
```
Turtlelord26: Suggest phrasing: "all ^members^ of one team"

01_02_Combat_Rules.txt
```diff
@@ -999,3 +999,128 @@ retrieve a dropped weapon or stand back up if they have fallen.
...
+longer fight or the breaching order has been exhausted. Play resumes as normal
+after the turn of the last character to stack up after that.
+
+===== Interrupting a Stackup =====
+
+    If an enemy approaches or attacks one of the characters in a stack up line
+after they have stacked up but before the lead character has called "breach!",
+then all of the characters in the stack up are no longer considered to be 
+stacked up. Instead, they may use their held actions as immediate reactions in
+initative order and must wait until their next turn to stack back up.
```
Turtlelord26: All of those held actions? I wonder whether there should there be some penalty for being interrupted while stacking, perhaps you only get one Reaction, or one reaction plus a cover dive like in the start of combat rules. This point is debateable.

01_02_Combat_Rules.txt
```diff
@@ -999,3 +999,128 @@ retrieve a dropped weapon or stand back up if they have fallen.
...
+disadvantage and the attackers roll reflex with advantage.
+
+*Breaching Roll = Reflex Save + Weapon Reflex Mod + any mod from feats
+
+===== Clearing =====
+
+    Players take their turn in decending order of breaching roll. In other words,
+The character with the highest net reflex roll goes first, the character with 
+the lowest reflex goes last. Characters on the attacking team (the team that 
+stacked up) must use an Action to move from their position in the stack line
```
Turtlelord26, line 1100: It's been awkward during room clears when you're fourth in line but roll higher in the ordering than the player in first. Do they jump the line or wait and squander that glorious reflex roll? 
It's probably been as bad when the defenders roll higher than the attackers and can't shoot anything because nobody's entered the room yet.

Would it be feasible to stick an Entry phase between Breach and Clear, where the attackers all execute a movement action in stackup order to enter the room? I think it would reduce the purely mechanical obnoxiousness that linear action order imposes and also make these events even more bloodily (viscerally?) quick, which I think would feel authentic and novel.

01_02_Combat_Rules.txt
```diff
@@ -999,3 +999,128 @@ retrieve a dropped weapon or stand back up if they have fallen.
...
+entry point from where your character is stacking up. On the rare occasion that
+one door leads to more than one room, a single room must be designated for 
+clearing. You cannot breach and clear a room that is not attached to the entry
+point that you stacked up on. 
+
+    If your turn starts and you are already stacked up, you may either leave the
+line of the stack up and take your turn normally, or skip your turn to remain in
+the stack up line. If you skip your turn in this manner, you will have the same
+number of actions during clearing as you would if you stacked up as the first
+act of your turn.
```
Turtlelord26: It may not be clear to everyone whether staying in a stackup line is an Action or a Free Action.

01_02_Combat_Rules.txt
```diff
@@ -999,3 +999,128 @@ retrieve a dropped weapon or stand back up if they have fallen.
...
+    Any number of characters can stack up on a given entry point. The order of
+the characters in the lineup is important. Players can chose any position in the
+stackup order that hasn't yet been taken at the time that they declare their 
+character will stack up. So the first character to stack up could choose to get
+in the middle of the line, and the second character to use their turn to stack
+up could stack up in front if they so chose.  
+
+===== Breaching =====
+
+    Any time a character joins the stack up line, the character in the lead of
```
Turtlelord26, line 1072: Brilliant use of a very specific and highly relevant trigger condition. The 🐢 approves.

Turtlelord26 (changes requested):
I love that you built this to allow for smooth use during a larger battle. I'm slightly worried about the  Shahrazad effect (I can explain this reference in Slack; the gist is recursive subgames are dangerous to game flow), but the only way to find out about that is to play it and see how it feels and tracks.

I also love writing down things we've been houseruling the whole time, and can't wait to get some real balance feedback on weapon reflex now that we'll actually use it.


This is bleedingly close to shippable, but I want at least a short conversation on my line 1098 comment before greenlighting. Most of the rest of the notes are nonblocking improvements.

01_02_Combat_Rules.txt
bleehu: I was wondering that too as I wrote it, but was racing to get it all down so we had a complete system to iterate on. I like the one reaction + cover dive. I'll do that.

01_02_Combat_Rules.txt
bleehu: thanks! :)

01_02_Combat_Rules.txt
bleehu: In the defender-high position, they simply hold actions and blast the door or try to shoot through walls. The way I have it written, you have the option of jumping line and moving through your entry point in reflex order, as long as someone isn't already blocking it. My hesitation to add an Entry phase is that I think some breaches don't necessarily involve physical entry of the room until the shooting has stopped. It may be that waste of reflex is a mechanical penalty for trying to squeeze your dudes through a small hole. I'll do a little more research and come back to this. I'm still considering adding the Entry Phase as you described it.

Turtlelord26 (approved):
I'm good with leaving entry phase speculation to a new issue where we can consider it without blocking the rest.
Would still lean toward a 1 condition = 1 (dis)advantage rule of thumb, but I can wait and see how it plays.
Shipit the things!

bleehu:
You read my mind!

Issue #621, opened 2018-07-31
Nanite Removal Issue Tracking: Rests


Turtlelord26:
Working on the Turtlelord26-Nanite_Removal_Project-Superbranch

Removed nanite mention from short and long rests, leaving them with no effects. They need effects.

Rests are known to heal characters and recharge battery packs, but may do more things.

PR #744, opened 2018-12-30
Add rules text to Rests!

Turtlelord26:
#621
We have definitions, conditions, and benefits!
Slight production sin in Long Rest where I refer to "Exhaustion Rules" that don't exist yet. Mea Culpa.


Reveraine (approved):
These are wonderful!

bleehu (approved):
I like it! Shipit!

PR #752, opened 2019-01-10
Wording Minutiae

Turtlelord26:
No one rules lawyers their way into shenanigans because we neglect to specify timings.


bleehu (approved):
Good catch!!!

Issue #759, opened 2019-01-13
Bug: No Blindness Condition


rosebrallier:
Blindness is referred to in Flash Bangs and in Human Species files, but is not described under Conditions.

bleehu:
Good Catch, R!

Issue #758, opened 2019-01-13
Gripe: Thermal damage missing


rosebrallier:
I thought we agreed to roll Cold and Fire damage into Thermal damage, but there is no Thermal (or Cold) damage in [Damage and Armor Types](https://github.com/bleehu/Compound_X/blob/dev/01_04_Damage_Types_and_Armor_Types.txt), only Fire damage. When did this happen? Was it on purpose?
Blocking editing work on Human Species.

bleehu:
I think I got called out. I failed to make the change in this PR. https://github.com/bleehu/Compound_X/pull/358/files 

That would mean that our frost-resistant Western Earthlings would be resistant to fire as well as frost, but I think that actually holds more accurately to how physics works. 
https://github.com/bleehu/Compound_X/blob/Races-Push/CharacterCreation/01_01_Species_Human.txt#L171

bleehu:
I'll tackle this as part of gripe processing in sprint-after-next.

PR #760, opened 2019-01-13
Humans Patch

rosebrallier:
Made a variety of clarifying edits to Humans.
Remaining edits required:
- [ ] line 42: update comparative size difference between Human, Opaleite and Tundari Empires (James says we decided that the Opaleite Empire was much closer in size to the Human)
- [x] line 125: Sixth Sense functionality requires mechanical clarification @bleehu 
- [ ] line 133: Imo, Eastern and Western Earthlings should live in the Eastern and Western Hemispheres, since there is no East or West of the equator, only North and South.
- [ ] misc: Issues #758 and #759 created to address lack of Thermal Damage and Blindness Condition in Dev

### DIFF
--- a/CharacterCreation/01_01_Species_Human.txt
+++ b/CharacterCreation/01_01_Species_Human.txt
@@ -113,7 +113,8 @@ the 24th Century, forming a new colony in the Keplar system.
 
 Very Normal: 
 	Blending in has always come naturally to you. People tend to trust you. 
-You seem to belong no matter where you find yourself. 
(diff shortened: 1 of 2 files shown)


rosebrallier:
Updated to clarify Very Normal mechanics and remove Very Normal Feat

rosebrallier:
Proposed mechanics for Sixth Sense and removed the feat by the same name re Issue #772

bleehu (approved):
Great work RoseB! :shipit: !

Issue #620, opened 2018-07-31
Nanite Removal Issue Tracking: Active Feats


Turtlelord26:
Working on the Turtlelord26-Nanite_Removal_Project-Superbranch

Active feats need a dedicated rule paragraph to describe how to use Medical Nanites to pay for them.

bleehu:
This new rule should be in the feats document, right?

Turtlelord26:
Much like with the items documents, I would prefer feats to remain a pure list and leave the rules in the dedicated rules documents, in this case Basic Rules.

PR #784, opened 2019-01-14
Rules Text for Active Feats, now with Medical Nanites

Turtlelord26:
#620 
I've moved some work that was done in dev into this branch (namely, the movement of feat rules out of feat documents and into basic rules). I redid the active feat section to add medinites, and adjusted the prose of the other sections with intent to smoothen the read without functionally changing anything.

For reviewers: I recommend looking at the individual commit changes to more easily determine what I've changed from the big chunk I pulled in from dev (instead of the mass of green you'll see from the aggregate changes).

Notes
- I moved the feat rules downward from where I found them in dev, to after the skill rules. It seemed to me a natural progression that way, but is absolutely open for debate.


Reveraine (approved):
Lovely! This is a very good update to the feats rules.

PR #785, opened 2019-01-14
Nanite Aesthetic terminology in Disrupted

Turtlelord26:
Convert "nanite abilities" to "active feats", since those are the only nanite abilities left once the nanite superbranch merges. I'm making this change in dev instead of there because the updated flavor text postdates the superbranch's creation and it has enough merge conflicts already.


bleehu (approved):
:shipit: shipit!

Issue #721, opened 2018-11-28
Nanite Absorbsion Feat makes no sense


bleehu:
The nanite absorbsion feat was one of the last feats added before @rosebrallier and @Reveraine started keeping us honest about making our mechanics make some sort of sense. As written, Nanite absorbsion is buckshot [autocorrected] nanite magic. I'd _like_ to have something to replace this, but regardless of wether we have something new to put in, the old feat needs to come out.

PR #786, opened 2019-01-15
Delete Nanite Absorption

Turtlelord26:
per #721 
I can think of no way to keep nanite siphoning in-aesthetic.


bleehu (approved):
Shipit! :shipit:

bleehu:
Don't forget to mark Issue #721

PR #787, opened 2019-01-15
Delete Nanite Concentration

Turtlelord26:
While this one is more possible to keep in-aesthetic, I find the idea of using a feat to enhance the effects of other active feats to be of questionable value and a balance nightmare.


bleehu (approved):
nuke it from orbit! Shipit!

Issue #719, opened 2018-11-28
Focus Energy Feat uses outdated language


bleehu:
The "Focus Energy" feat is a holdover from the TRowl days, and still refers to "spells". My vote would be to just yank it from the game, but if someone can update it well, then we'd still have one more feat to chose from.

bleehu:
Turtle is fixing this in PR #788

PR #788, opened 2019-01-15
Delete Focus Energy

Turtlelord26:
per #719 
This is very space magic, and like nanite concentration, a balance nightmare. I like deleting it.
We can write new feats that know how nanites work now.


bleehu (approved):
Shipit

Issue #777, opened 2019-01-13
Clarify Tough Skinned Species Trait


bleehu:
The Opalite Chondrich subspecies has the "Tough Skined" trait. It could use some better prose, better mechanical description.

PR #793, opened 2019-01-16
added prose for the Condrich's Tough Skin ability

bleehu:
On branch MH/Issue-777/tough-skin
	modified:   01_04_Species_Opaleites.txt

Should close out Issue #777

### DIFF
--- a/CharacterCreation/01_04_Species_Opaleites.txt
+++ b/CharacterCreation/01_04_Species_Opaleites.txt
@@ -139,10 +139,20 @@ Appearance:
 
 == Available Traits ==
 
-Tough Skinned
-    
-    [+1 to armor range while being attacked by a melee weapon]
-    
+Tough Skinned:
+
+    The chitinous interlocking scales of the Condrichs are surprisingly adept at
+naturally dispersing the impact from bullets, toothy predators and melee weapons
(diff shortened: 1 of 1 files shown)


Turtlelord26 (changes requested):
This isn't just prose, you rewrote the mechanics too. I like the concept a lot, but giving an armor seems maybe too similar to the racial armor trait drones get. 
Also, the plus 40 AS alarms me - it nearly doubles any <5 level character's damage soak per turn. Not in line with other racial traits' strength, and it could make chondrich mandatory for tanks.

rosebrallier:
@Turtlelord26 Personally, I think the APL of 1 somewhat (though not entirely) balances the rest of the trait's strength, but you have a point overall; the +40 AS should probably be nerfed.

bleehu:
In my mind, 40 AS is one pistol round, one SMG round or half an assault rifle round. That seems sane to me. If it still seems too high, I'll nerf it.

Reveraine (commented):
I think that this change is good, but probably needs to be balanced further. 
I think either the coverage amount goes down, or the armor strength does (or it needs to have no per-round regeneration, but that sounds needlessly complex). In any case, I think it is close to ready.

Turtlelord26 (approved):

Issue #619, opened 2018-07-31
Nanite Removal Issue Tracking: Medical Nanite Item


Turtlelord26:
Working on the Turtlelord26-Nanite_Removal_Project-Superbranch

Added a dummy item for Medical nanites. They need their description and rules filled out.
In particular, they need a unit limit by level rule, possibly as a new column in the leveling table.

PR #802, opened 2019-01-31
Textify Medical Nanites

Turtlelord26:
#619 
They now have built-in benefits (minor Shock save bonus added to the already-existing rest healing), and link you to relevant rest and feat rules.
Provided explicit pathway to upgrade from the booster to the full unit.


Items/07_Items.txt
```diff
@@ -38,12 +38,25 @@ inside it, exportable to a portable device.
...
+may be upgraded to a full unit (see below) at a cost of $700.
+	
+Medical Nanites Unit ($1000):
+	One unit of Medical Nanites. Having at least one unit of Medical Nanites 
+provides all the benefits of a booster shot (see above), with even more potent 
+rest healing. Additionally, full units of medical nanites can be harnessed to 
+execute Active Feats (see Feats in the Basic Rules).
+	Additional units after the first do not provide more healing or save 
+bonus, but may be used to execute more active feats in shorter time frames. The 
+number of Units a character can have at once is related to their level, see the
```
rosebrallier: We could say "maximum number of Units" to be as super clear as possible.

Items/07_Items.txt
```diff
@@ -38,12 +38,25 @@ inside it, exportable to a portable device.
...
+	Additional boosters after the first have no effect. A booster shot 
+may be upgraded to a full unit (see below) at a cost of $700.
+	
+Medical Nanites Unit ($1000):
+	One unit of Medical Nanites. Having at least one unit of Medical Nanites 
+provides all the benefits of a booster shot (see above), with even more potent 
+rest healing. Additionally, full units of medical nanites can be harnessed to 
+execute Active Feats (see Feats in the Basic Rules).
+	Additional units after the first do not provide more healing or save 
+bonus, but may be used to execute more active feats in shorter time frames. The
```
rosebrallier: "provide more save bonus" reads really awk to me. Perhaps "do not provide more healing over rests or increase the save bonus"

Items/07_Items.txt
```diff
@@ -38,12 +38,25 @@ inside it, exportable to a portable device.
...
-Medical Nanites Booster Shot ($):
-	Half a unit of Medical Nanites.
+Medical Nanites Booster Shot ($400):
+	A partial unit of Medical Nanites. The booster shot is a standard 
+medical procedure on more affluent worlds, like a vaccine, and provides 
+resistance to and faster healing from disease and injury. Characters with a 
+nanite booster shot gain +10 to Shock saves and benefit from faster healing 
+during Short Rests (see Resting in the Basic Rules).
+	Additional boosters after the first have no effect. A booster shot 
+may be upgraded to a full unit (see below) at a cost of $700.
```
rosebrallier: Suggestion for the Mostest Clarity: "upgraded to a Medical Nanites Unit"

rosebrallier (approved):
Yay elaborate the thing!
I left two clarity suggestions and one more critical lang style request.

01_06_Leveling_Up.txt
```diff
@@ -15,48 +15,48 @@ skill points for that level.
...
-| 16    | "        | 9     |            |            | 13            | "                  |
-+-------+----------+-------+------------+------------+---------------+--------------------+
-| 17    | "        |       |            |            |               | "                  |
-+-------+----------+-------+------------+------------+---------------+--------------------+
-| 18    | "        | 10    |            |            | 14            | 7d10               |
-+-------+----------+-------+------------+------------+---------------+--------------------+
-| 19    | "        |       |            |            |               | "                  |
-+-------+----------+-------+------------+------------+---------------+--------------------+
-| 20    | "        | 11    | 4          | +1         | 16            | "                  |
-+-------+----------+-------+------------+------------+---------------+--------------------+
```
bleehu, line 62: That's a reward that is going to feel very good! It's worth calling out that this is going to have major balance ramifications on enemy levels, but I think this is ready to test. Looking forward to it!

Items/07_Items.txt
```diff
@@ -32,18 +32,31 @@ Engineer processes; i.e., a Gunslinger who buys an extruder can't construct
...
+	
+Medical Nanites Unit ($1000):
+	One unit of Medical Nanites. Having at least one unit of Medical Nanites 
+provides all the benefits of a booster shot (see above), with even more potent 
+rest healing. Additionally, full units of medical nanites can be harnessed to 
+execute Active Feats (see Feats in the Basic Rules).
+	Additional units after the first do not provide more healing or increase 
+the save bonus, but may be used to execute more active feats in shorter time 
+frames. The maximum number of Units a character can have at once is related to 
+their level - see the leveling table.
```
bleehu, line 59: Blocker. This is a really good idea, but I think we need to add edge-case guard rails. In formal terms, this allows the linear-scale buying of power without depreciating marginal gains. Or in laymen's terms, what happens if someone sells a small spaceship and dumps all of that money into Med Nanites. 
One potential guard rail would be a by-level maximum which is explained diajetically that introducing too many nanites to your system at once is really hard on your immune system. Another guard rail could be a limit to only being able to get one more unit per level.

bleehu (changes requested):
Need a guard rail to prevent out-of-control nanite unit abuse. Once that's done, this will be ready! Great work on the mechanic itself though! Prices all look good, and I think this will be a fabulous reward for players as they level up.

Items/07_Items.txt
Turtlelord26: I hadn't considered waiting until level 8 and dumping three units into yourself all at once. I would want to add acquisition rate-limits based on flavor alone.
Are you also asking me to make units after the first cost more?

bleehu (approved):
Great flavor! I can work with that.

Issue #694, opened 2018-09-16
Condition- Unconscious


rosebrallier:
Oooops.

rosebrallier:
Closed re PR #808

PR #806, opened 2019-02-01
Unconsciousness wheeeeeee

rosebrallier:
Now we have a condition to reference that is "drop like a rock w/o necessarily being @ 0 hp"


rosebrallier:
Re Issue #694

PR #808, opened 2019-02-01
Unconscious condition

rosebrallier:
Also updated blacking out to result in Unconsciousness rather than Stun re Issue #694


Turtlelord26 (commented):
Can we also mention that the unconscious character is unaware of their surroundings?

bleehu (approved):
Agree with :turtle:  on unaware, would also like to add unable to speak. But I also feel that is very strongly implied by the title of the condition.

rosebrallier:
> also like to add unable to speak. But I also feel that is very strongly implied by the title of the 
condition.

I consider speaking to be included under "can take no actions"- Free actions are still actions.

Issue #776, opened 2019-01-13
Add balanced numbers to "Powerful" Species Trait


bleehu:
The Species trait "Powerful" in the Sang'Mau Tundari Species doc could use some specific mechanics.

https://github.com/bleehu/Compound_X/blob/Races-Push/CharacterCreation/01_03_Species_Tundari.txt#L201

PR #819, opened 2019-03-02
Add numbers to the Sang'mau Powerful trait

Turtlelord26:
#776 
I wrote down my reasoning in the commit description 🐢 

Edit: forgot to mention: I also ran the numbers with +10 damage at lv 1, and it sped up the assassin kill by 2 turns instead of 1, while the zerker kill speed was still 1 turn faster. The overkill cushion was much more comfortable in both cases.

---------------------------------------

It's interesting to me because I don't think you can make a character that really benefits from both parts of this racial. I decided to give it initial numbers based on only one of them being truly relevant to a given character.

The other amusing thing is that giving carry ability lets heavy gunners skimp a bit on strength or fort, but they need a fair amount of strength for the gun stat requirements and can't really get around that. Therefore I expect the average Powerful character to have lower toughness than non-Powerful characters.

The 5 damage to unarmed strikes is specifically lower than the bonus from taking unarmed combat feats, since those are rate-limited to every 5 levels and jumping that line is quite powerful. 
For the math (characters with 9/7 STR/DEX)
a level 1 Assassin does 44 damage per unarmed strike
a level 1 Berserker does 64 damage per unarmed strike by virtue of starting out with Unarmed Combat training.
a level 15 Assassin/Berserker who put all their stat increases into STR/DEX and took all the unarmed combat feats would deal 136 damage per unarmed strike. _Yes, you read that right. Check your corners._
The takeaway is that a flat bonus disproportionately aids assassins specifically at level 1, but a percentage increase is just too terrifying in the lategame. I decided to go with flat even though the flavor-mechanics interaction is a little unfortunate. Finally, the flat bonus increases over time to keep it relevant, as otherwise it would drop off fast.
Effect: 
at lv 1 vs an enemy with 120 health and 40 AP that the melee character never bypassas (hah), giving 2 successful attacks per turn:
Assassin kills in 6 attacks, overkill by 24 HP. Powerful Assassin kills in 5 attacks, overkill by 5 HP.
Berserker kills in 4 attacks, overkill by 56 HP. Powerful Berserker kills in 3 attacks, overkill by 7 HP

I like these attack numbers here, and the overkill margin being so small means we'll have relatively dramatic testing results.

### DIFF
--- a/CharacterCreation/01_03_Species_Tundari.txt
+++ b/CharacterCreation/01_03_Species_Tundari.txt
@@ -207,7 +207,7 @@ Climbing:
 
 Powerful: 
 	Large, developed muscles allow you to carry more and hit harder. 
-[Increase Carry Ability by {Vaguely balanced number}. Increase damage of Unarmed Strike by {Vaguely balanced number}.]
+[Increase your Carry Ability by 2, and increase the damage of your Unarmed Strike by 5. At levels 5, 9, 15, and 20, further increase the damage of your Unarmed Strike by 5.]
 
 
 


bleehu (approved):
Great write up! I would have thought this was a bit low, but you've got me very well convinced!

bleehu:
And no kidding, check your corners!

Issue #671, opened 2018-09-09
Gripe: How much ammo do I get to start?


bleehu:
We should make this explicit, perhaps in a new Making a character doc?

PR #822, opened 2019-03-04
added rules on buying and selling.

bleehu:
Added rules about ammo on weapon purchase
	modified:   01_01_Basic_Rules.txt
Should finally address issue #671

### DIFF
--- a/01_01_Basic_Rules.txt
+++ b/01_01_Basic_Rules.txt
@@ -631,18 +631,18 @@ Items and Buying Items
 ==============================
 
 	When you build a character, you will get some starting money and the time to
-spend it to acrue the kit that you'll start playing with. But this likely won't
-be the last time that you'll have the opportunity to buy things. It may be smart
-to save some of your starting cash to spend elsewhere. Some stores and vendors
-you meet will sell exactly what you see in the Player Handbook. Some traders 
-may offer items or weapons that aren't available either at the start of the game
-or anywhere else in the galaxy! 
+spend it to accrue the kit that you'll start playing with. However this likely 
+won't be the last time that you'll have the opportunity to buy things. It may be 
+smart to save some of your starting cash to spend elsewhere. Some stores and 
+vendors you meet will sell exactly what you see in the Player Handbook. Some 
+traders may offer items or weapons that aren't available either at the start of 
+the game or anywhere else in the galaxy! 
 
 	When you buy a gun, the gun usually comes with 5 magazines of ammunition
-unless otherwise stated. If you spend an Action to pick up a weapon from a 
-fallen enemy or comrade, that weapon will only have the magazine that was in it
-when it dropped. However, most enemies will carry more ammo into battle, and if
-you scavenge around, you may get lucky!
+unless otherwise stated. If you pick up a weapon from a fallen enemy or comrade, 
+that weapon will only have the magazine that was in it when it dropped. However, 
+most enemies will carry more ammo into battle, and if you scavenge around, you 
+may get lucky!
 
 	All prices listed in the Player Handbook are what you will pay when making a
 character. Once your adventure has begun however, a trader may adjust their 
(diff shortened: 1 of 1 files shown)


01_01_Basic_Rules.txt
```diff
@@ -628,6 +626,33 @@ Class Feats:
 specifies a class feat. Characters can not take non-class feats instead of class feats 
 or vice-versa. Class Feats are listed in the Classes document.
 
+==============================
+Items and Buying Items
+==============================
+
+	When you build a character, you will get some starting money and the time to
+spend it to acrue the kit that you'll start playing with. But this likely won't
```
Turtlelord26: sp: accrue

01_01_Basic_Rules.txt
```diff
@@ -628,6 +626,33 @@ Class Feats:
...
+be the last time that you'll have the opportunity to buy things. It may be smart
+to save some of your starting cash to spend elsewhere. Some stores and vendors
+you meet will sell exactly what you see in the Player Handbook. Some traders 
+may offer items or weapons that aren't available either at the start of the game
+or anywhere else in the galaxy! 
+
+	When you buy a gun, the gun usually comes with 5 magazines of ammunition
```
Turtlelord26, line 641: 5 is many, since we've usually been playing with 2 or 3. Any particular reason?

01_01_Basic_Rules.txt
```diff
@@ -628,6 +626,33 @@ Class Feats:
...
+unless otherwise stated. If you spend an Action to pick up a weapon from a
```
Turtlelord26: I would leave out the 'spend an Action'. It will almost always cost an Action to pick up an item, but if it did eventually change, we'd change it in the action rules and never remember to change it here too. DRY it out please 😃

Turtlelord26 (approved):
I like it. One question and two nonblockers below.

01_01_Basic_Rules.txt
rosebrallier: I'd replace "But" with "However," for a more structurally complete sentence. (The "but" sticks out in a bad way here imo)

01_01_Basic_Rules.txt
```diff
@@ -628,6 +626,33 @@ Class Feats:
...
+you scavenge around, you may get lucky!
+
+	All prices listed in the Player Handbook are what you will pay when making a
+character. Once your adventure has begun however, a trader may adjust their 
+prices on you. Most traders in the Compound X universe make their living by 
+buying low and selling high. These merchants might raise their prices above the
+going rate if they don't like you, or if they think you're dumb. On the other 
+hand, if you're particularly charming or if you just saved their town, they may
+offer you discounts. Many, but not all characters can be haggled with if you
+have a high Charisma. As always, what the GM says goes.
```
rosebrallier: I'd like to note that having high Cha has no bearing on whether a pc _can_ haggle with and npc, just on whether they can do so _successfully._

01_01_Basic_Rules.txt
```diff
@@ -628,6 +626,33 @@ Class Feats:
...
+unless otherwise stated. If you spend an Action to pick up a weapon from a 
+fallen enemy or comrade, that weapon will only have the magazine that was in it
+when it dropped. However, most enemies will carry more ammo into battle, and if
+you scavenge around, you may get lucky!
+
+	All prices listed in the Player Handbook are what you will pay when making a
+character. Once your adventure has begun however, a trader may adjust their 
+prices on you. Most traders in the Compound X universe make their living by 
+buying low and selling high. These merchants might raise their prices above the
+going rate if they don't like you, or if they think you're dumb. On the other
```
rosebrallier, line 651: Plus, the going rate won't be the same across the galaxy. Snow's a lot less expensive in Antarctica than in Fiji.

rosebrallier (approved):
I'm in pending Turtle's question! Take the language notes under advisement.

01_01_Basic_Rules.txt
bleehu: That's what 1Source had as his default at one point? I'd like players not to be so ammo-conscious that they abstain from suppression. But my heart isn't set on that number by any means.

01_01_Basic_Rules.txt
bleehu: oh snap! With the right acronym and everything! You're absolutely right!

01_01_Basic_Rules.txt
Turtlelord26: Is that a problem we can solve instead by separating costs for magazines of ammo and bulk ammo (with competitiveTM bulk pricing)?

01_01_Basic_Rules.txt
bleehu: ya, I really shoudn't start a sentence with "but".

01_01_Basic_Rules.txt
bleehu: Very good distinction!

PR #826, opened 2019-03-04
Add mechanics to Tundari Clawed trait

Turtlelord26:
Instead of more raw damage, which would stack unacceptably with Powerful down in Sang'mau, I thought I would try something slightly different.

It may still be too synergistic with Powerful, and that should be tested. If that is the case, another option is moving Powerful's damage bonus here and redoing Powerful as the carry bonus plus a bonus to some str-based skill or maneuver check.


bleehu (approved):
I'm a little concerned that it's underpowered, but let's give it a shot!

Issue #810, opened 2019-02-04
Gripe: Improvised Thrown Weapons


Reveraine:
Improvised thrown weapons needs guard rules. 
The possible distance for Improvised thrown weapons needs to be reevaluated.

rosebrallier:
So, a strict reading of the [thrown weapons rules](https://github.com/bleehu/Compound_X/blob/f1201dd0744a03568d550640ef6aa60bf5f27885/01_02_Combat_Rules.txt#L392) indicates that Improvised Thrown Weapons should have Miss Chances equal to a miscellaneous Thrown Weapon of the same size, [-2 Accuracy.](https://github.com/bleehu/Compound_X/blob/f1201dd0744a03568d550640ef6aa60bf5f27885/01_02_Combat_Rules.txt#L237)
So, they don't need Guard rules, but they do need entries in [Melee and Thrown Weapons](https://github.com/bleehu/Compound_X/blob/dev/Items/04_01_Melee_%26_Thrown_Weapons.txt ) (one per weapon size).

PR #828, opened 2019-03-06
Add Miss Chance to Improvised Weapons for Throwing

Turtlelord26:
per #810

### DIFF
--- a/Items/04_01_Melee_&_Thrown_Weapons.txt
+++ b/Items/04_01_Melee_&_Thrown_Weapons.txt
@@ -21,8 +21,14 @@ Damage: 30 + STR mod + DEX mod Ballistic, minimum 10
 Range: Adjacent (1m)
 Accuracy Modifier: 0
 
+	Ranged Miss Chance
+
+* 2 out to 5m
+* 5 out to 20m
+
 	Some improvised weapons may have alternate damage types, at the DM's 
-discretion.
+discretion. Note that the miss chance stat is only relevant when throwing an 
+improvised weapon.
 
 == Improvised One-handed Weapon ==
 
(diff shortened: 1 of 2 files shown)


Items/04_01_Melee_&_Thrown_Weapons.txt
```diff
@@ -21,26 +21,45 @@ Damage: 30 + STR mod + DEX mod Ballistic, minimum 10
 Range: Adjacent (1m)
 Accuracy Modifier: 0
 
+	Ranged Miss Chance
+
+* 2 out to 5m
+* 5 out to 20m
```
bleehu, line 27: did we want to add strength scaling for how far a thrown improvised weapon will go?

Items/04_01_Melee_&_Thrown_Weapons.txt
```diff
@@ -21,26 +21,45 @@ Damage: 30 + STR mod + DEX mod Ballistic, minimum 10
...
 
+	Ranged Miss Chance
+
+* 2 out to 5m
+* 5 out to 20m
+
 	Some improvised weapons may have alternate damage types, at the DM's 
-discretion.
+discretion. Note that the miss chance stat is only relevant when throwing an 
+improvised weapon, while the accuracy is only used for melee attacks.
```
bleehu, line 31: great clarification! I love how I can understand these rules well the first time I read them. That wasn't always true with CX rules.

bleehu (approved):
Some minor suggested tweaks, but this would play very well!

Items/04_01_Melee_&_Thrown_Weapons.txt
Turtlelord26: We could, but I have a suspicion that most of the throwing weapon fantasy is associated with dexy-assassin-type characters, and so making throwing weapons' full potential rely on strength would feel bad. We already use strength scaling on the damage, and that satisfies me on the flavor front.

PR #830, opened 2019-03-07
Added Subspecies Blurbs

dfuchs2:
Created short explanations of the various subspecies.


bleehu (approved):
I like it! Might want to add a little more detail on the mechanics of a sub-species (i.e., you can't have more than one subspecies at a time, you choose one sub-species on character creation, etc, etc) but we can add that later if need be.

Issue #783, opened 2019-01-14
Thrown Weapon Damage from Unarmed Feats


Turtlelord26:
https://github.com/bleehu/Compound_X/blob/Turtlelord26-Nanite_Removal_Project-Superbranch/01_02_Combat_Rules.txt#L343

This line refers to a damage bonus granted to thrown weapons by the unarmed combat feats. I really should cull it and put the damage bonus in the feat descriptions or no one will remember this rule.

PR #831, opened 2019-03-13
Delete vestigial thrown weapon rules text

Turtlelord26:
#783 + scope creep
Removed damage boost from unarmed combat feats - off flavor and we can adjust weapon damages by level requirement if we have to.
Removed skill points reducing attribute requirements - unnecessary complexity not mirrored in any other weapon.
Removed blanket Reflex modifier statement - each weapon should have a reflex modifier stated as the weapon statblock is where players will look for that information.
Removed outdated reference to attributes affecting range.


bleehu (approved):
Much easier to read!

PR #832, opened 2019-03-24
Minor rework of Twitch feat

Turtlelord26:
Now a Reaction instead of a Free Action. Also now has fewer bizarre edge cases for lingering accuracy penalties.


bleehu (approved):
Shipit!

Issue #695, opened 2018-09-16
Vehicle master class feat unbalanced


bleehu:
The vehicle master Pilot Class feat was written before our new and improved skill rules. It allows a pilot to roll over 200; the hard limit. It needs to be nerfed or replaced.

PR #852, opened 2019-04-12
Straight Deletion of Vehicle Mastery Pilot Feats

Turtlelord26:
#695 
Pilots had 6, now down to 4, which is similar to many other classes.


bleehu (approved):
Not really sure why we want to remove content that isn't broken, but If you can give a good reason for it, yank it.

Turtlelord26:
> Vehicle master class feat unbalanced #695
> bleehu opened this issue on Sep 16, 2018 · 0 comments
> The vehicle master Pilot Class feat was written before our new and improved skill rules. It allows a pilot to roll over 200; the hard limit. It needs to be nerfed or replaced.

Issue #796, opened 2019-01-21
Damage types are out of order


bleehu:
What does "misc damage types" mean anyway? Why is fall damage under protection types? We should reorganize that whole file.

Turtlelord26:
Split into #866 and #868

PR #866, opened 2019-05-24
Reorder damage types.

Turtlelord26:
#796 partially

* Reordered damage types and renamed categories. Common damage types are common, Additional damage types are the rarer ones.
* Alphabetized the Additional damage types.
* Rewrote some prose for armor and shield blurbs and added a summary of damage types they interact with for quick reference.
* Turned fall damage into its own section instead of a subsection, but it really should move out of this file and into basic rules (or a subdocument of basic rules labeled Adventuring with things like Size, Light levels, Noise levels, and Carrying Weight rules) @bleehu did I just describe a new issue?


bleehu:
issued https://github.com/bleehu/Compound_X/issues/868

bleehu (approved):
shipit!
Looks good!

PR #874, opened 2019-06-13
added flavor text to military background.

bleehu:


bleehu:
Should there have been more to do here?

rosebrallier (commented):
This is a great start. Personally, I'd like the flavor text to be a little more dramatic like, "When you're shot at, the fear and adrenaline still pulse through you, but then your training kicks in."
OOOOOO We could re-name it to Discipline!!!!!

bleehu:
It is so.

rosebrallier (approved):
Wooo!

Issue #846, opened 2019-04-12
Write Academic Background Perks


rosebrallier:
Requires description for Expert Researcher and mechanics and description for Birds of a Feather

rosebrallier:
PR into [rosebrallier-backgrounds](https://github.com/bleehu/Compound_X/tree/rosebrallier-backgrounds) --> CharacterCreation --> [04_Backgrounds](https://github.com/bleehu/Compound_X/blob/rosebrallier-backgrounds/CharacterCreation/04_Backgrounds.txt)

PR #876, opened 2019-06-14
added background 'I minored in that'

bleehu:

What do you think of that?


bleehu:
Should wrap up the latter half of Issue #846

rosebrallier (approved):
Shipit!

PR #879, opened 2019-06-18
Bumped version number to 0.3.0

Turtlelord26:
A major system rework in species merits an elevated version escalation, methinks.
Last time was the melee rework and this is bigger 🐢


bleehu (approved):
Shipit!

PR #881, opened 2019-06-19
Mh/3.0fixes/endurance

bleehu:
I'm not sure if this is necessary. Free sprinting each time you sprint is nice, but not game breaking.

### DIFF
--- a/CharacterCreation/01_01_Species_Human.txt
+++ b/CharacterCreation/01_01_Species_Human.txt
@@ -145,8 +145,8 @@ Endurance:
 	Generations of Nomadic behavior have made moving long distances 
 easy for those of Eastern Earthling descent. 
 
-    Once per engagement, your first 2 Actions Sprinting are free of penalty. 1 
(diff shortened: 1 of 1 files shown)


Turtlelord26 (commented):
I'm not actually sure what's going on with this trait, on reflection. Normally, there's no mention of a Recovery action in the regular sprinting rules, so this is technically a negative trait?
There might be a more involved conversation we need to have here

Ref: https://github.com/bleehu/Compound_X/blob/dev/Combat%20Rules/1_Combat_Rules.txt#L126

Turtlelord26:
What do you think of "Once per engagement, your Sprinting actions during one turn do not incur the normal penalties." and just discarding the recovery action business?

Slight buff but removes the feel-bad of when you need 3 sprint actions to get to your target and the third one applies those penalties you were so close to avoiding.

bleehu:
Works for me. I'll update.

Turtlelord26 (approved):

PR #882, opened 2019-06-19
Mh/3.0fixes/d filter

bleehu:
The effect of the abilitiy is explicitly stated in the flavor text, but I can repeat it nonetheless.


Turtlelord26 (changes requested):
Rebase to dev please 🐢 

Flame retardant needs more than a 1-word change, since thermal damage can result from cold or hot sources. Not sure about stealth editing it here.

For the D-filter itself, suggest a reordering instead of a repetition. What would you think of 
> Due to the frequent ... in their Optic sensors. Unfortunately, this augmentation does take ... for such night vision.
> 
> [your new text]

Turtlelord26 (approved):

PR #885, opened 2019-06-19
changed functional text of hind-claw species feat

bleehu:
modified:   01_03_Species_Tundari.txt


Turtlelord26 (approved):
Good, now testable. Slight concern that +1 move per action is twice as powerful as pre-rework tundari +1 movement speed.

PR #886, opened 2019-06-19
removed climbing from Chi'mau

bleehu:
modified:   01_03_Species_Tundari.txt


Turtlelord26 (approved):
Correction: you removed it from Chi'mau. That's what I wanted though.

PR #887, opened 2019-06-21
Updated Functional Text: Toothed trait

Turtlelord26:


bleehu:
Why are we making this change?

Turtlelord26:
I didn't like the open-endedness of the wording "initial intimidation check".

On the new wording being open to applying the bonus multiple times to the same target within [a timeframe], I supposed that most DMs would rule that if you fail to intimidate someone, you can't try again immediately anyway, so it's not an enormous buff. The alternative added word count.

bleehu (approved):
I think big teeth should work on all biological characters; as a human, I'm just a little more intimidated by dudes with big biceps. But I don't think it's worth blocking.

PR #888, opened 2019-06-21
Specify climbing boundary in Climbing trait

Turtlelord26:


bleehu (changes requested):
Too specific; If you're making a climbing check, then you're making a climbing check. The degrees either doesn't matter or depends on the material you're climbing.

Turtlelord26:
You are correct, I'm trying to do something dubiously necessary in definitely the wrong spot.

PR #889, opened 2019-06-21
Name-change to Dark Vision trait

Turtlelord26:
Now no longer potentially confuses DnD players used to rolling up Drow PCs.


Turtlelord26:
Duplicated in #903

PR #890, opened 2019-06-21
Skilltext -> Swimmer trait

Turtlelord26:
Better templating. Additionally changed flat breath-holding text to an additive on the assumption we will eventually have rules for breath baseline, based on FORT.


bleehu (approved):
shipit!

PR #891, opened 2019-06-21
Wording update to Flame Retardant trait

Turtlelord26:
now references low- and high-temperature Thermal damage sources, and has more flexible wording on the environmental parts in anticipation of rules for that SoonTM.


bleehu (approved):
perfect. Much easier to read. Shipit!

Issue #840, opened 2019-04-12
Write Urchin Perks


rosebrallier:
Requires description for Below the Radar

rosebrallier:
PR into [rosebrallier-backgrounds](https://github.com/bleehu/Compound_X/tree/rosebrallier-backgrounds) --> CharacterCreation --> [04_Backgrounds](https://github.com/bleehu/Compound_X/blob/rosebrallier-backgrounds/CharacterCreation/04_Backgrounds.txt)

PR #896, opened 2019-06-22
added So You Say background trait

bleehu:

Issue #840 

I chafe just a little at the idea that a homeless person is necessarily sneaky or a pickpocket. We have a criminal background for that already. I propose that we give Urchins the gift of insight. In my experience, while a homeless person is more likely to have picked a pocket or hidden, they are all certain to have been lied to more than once. I'm not married to it as a change, it's just a proposal.


rosebrallier (approved):
This is a fantastic change! I love it! Personally, I'd merge the "but some lies" fragment into the previous sentence so that "Not that you'd show it" can stand out more.

Issue #845, opened 2019-04-12
Write Corporate Background Perks


rosebrallier:
Requires mechanics and description for Terms and Conditions and one complete perk including title, description, and mechanics

rosebrallier:
PR into [rosebrallier-backgrounds](https://github.com/bleehu/Compound_X/tree/rosebrallier-backgrounds) --> CharacterCreation --> [04_Backgrounds](https://github.com/bleehu/Compound_X/blob/rosebrallier-backgrounds/CharacterCreation/04_Backgrounds.txt)

PR #898, opened 2019-07-03
Corporate Background Perks

dfuchs2:
Added definition and mechanics for Terms and Conditions, and added My Way


dfuchs2:
https://github.com/bleehu/Compound_X/issues/845

rosebrallier (approved):
A+, very nice. We may need to up the Know-Corps bonus to 12 since it's a subskill, but balance isn't my thing, and it seems like a relatively useful skill in the CX setting. More info is required, so let's leave it as is.

PR #901, opened 2019-07-17
Rate-specify Wide-faced

Turtlelord26:
Largely a balance concern, partially because specifying a cooldown prevents one from being invented.


bleehu (approved):
Send it!

PR #902, opened 2019-07-17
Survivalist Trait re-skill

Turtlelord26:
Survival makes more sense to me given the clan's flavor. Also jury-rigging scares me.

Open to debate.


bleehu (approved):
I mean, it's litterally called "survivalist"...

PR #903, opened 2019-07-17
Rename Darkvision

Turtlelord26:


bleehu (approved):
Agreed. It's a little derivative.

PR #904, opened 2019-07-17
Removed Stat Gains From Opaleite Subspecies

Turtlelord26:
No other subspecies was doing it, and it made Opaleites concerningly powerful.


bleehu (approved):
I thought I'd already approved this?

PR #910, opened 2019-07-26
Even out Powerful trait scaling

Turtlelord26:


bleehu (approved):
Should we replace what's in the square brackets with more specific prose?

Turtlelord26:
It's a systematic templating inconsistency. Many earlierly written traits have formats like 
> Prose
> [Rules Text]

Standardizing the template is beyond the scope of this PR.

PR #911, opened 2019-07-26
Rework of Light Dance

Turtlelord26:
Subject to debate.

Removed "may be used to distract..." as that can be implied (so we don't need the extra text) and players will feel better about coming up with it on their own.
Removed that enemies may roll to negate the accuracy penalty over complexity and combat slowdown concerns.

Reworded ability to clarify what "while evading enemy fire" means. Now in effect if you moved at least your move speed on your previous turn. Slight nerf, hopefully makes up for removing the save.

### DIFF
--- a/CharacterCreation/01_04_Species_Opaleites.txt
+++ b/CharacterCreation/01_04_Species_Opaleites.txt
@@ -223,9 +223,9 @@ experienced individual can alternate lighting scales from tip to tail to
 misdirect an enemy.
 
    Under low-light or darker conditions, if you moved at least your move speed 
-on your last turn, enemies must pass a DC 60 Perception check in order to Aim
-at you (they may try multiple times). Enemies using nonvisible light, such as IR 
-goggles, are unaffected.
+on your last turn, then whenever an enemy would Aim at you, they must pass a 
+DC 60 Perception check to successfully do so. Enemies using nonvisible light 
+equipment, such as IR goggles, are unaffected.
    
 Deep Glow:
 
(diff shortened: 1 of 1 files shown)


bleehu (changes requested):
Requesting re-introduction of the DC 60 awareness to negate the effect; despite the added complexity, adds a nerf to the ability and a reward for aware enemies.

Turtlelord26:
Alright, willing to test.
Awareness save or Spotting check?

bleehu:
I think we use saves to prevent something bad from happening.

Turtlelord26:
In my reckoning, if the enemies have to roll when the ability is used (i.e. at the end of the Sea Devil's turn), then it's a save, but if the enemies have to roll when they try to attack the Sea Devil, it's a check, because they're taking action to overcome an obstacle (the hard-to-see property of their target).

It's less of a tracking headache for the DM to have the enemies roll when they attack, so they don't have to remember which enemies saw the Sea Devil light dancing.

To preserve the flatter roll distribution that saves have vs skills, it could be a straight Perception check instead.

bleehu:
That makes sense.

Turtlelord26:
I was thinking about how to phrase the perception check, and I had an idea. What if, instead of a -1 to accuracy unless they pass a check, enemies have to pass the check to Aim at the Sea Devil?

bleehu:
Very good idea, but I'm concerned it's too powerful.

Reveraine:
I agree that looking at it now makes it look a little strong. It still has to be in low-light conditions, but perhaps adding that a being wearing IR goggles will not be confused by this?

Turtlelord26:
I like the flavor of nonvisible light spectrum vision equipment bypassing Light Dance. I think it's acceptable mechanical complexity too.

bleehu (changes requested):
Should there be a rate limit or action cost on trying to use perception multiple times?

Turtlelord26:
The intention is that it's part of the Aim action. Will rephrase to make
that more obvious.

bleehu (approved):
Shipit!

PR #914, opened 2019-07-29
Power reduction on Engineering-minded trait

Turtlelord26:
I'm concerned about straight, unconditional advantage on specifically Jury-rigging but also in general for any skill.
Since we have this rolling the average technology, I propose starting testing at that power level.


Reveraine:
I like this change!

Reveraine (approved):

PR #921, opened 2019-08-10
Mh/issue 759/blindness condition

bleehu:
Added condition for blindness,
disadvantage on all skill checks, because being blind makes doing repairs hard and is distracting when you're trying to remember things.

should address gripe #759


bleehu:
aw heck. I need to rebase.

PR #922, opened 2019-08-10
added blind condition

bleehu:
modified:   01_05_Conditions.txt

Should address issue #759


bleehu:
Gives disadvantage on all skills because being blind makes doing repairs hard and remembering something when you've suddenly been blinded is distracting.

Turtlelord26:
Suggest also disadvantage on Reflex saves since we use Perception to calculate them.

Turtlelord26 (approved):
Shipit, sayeth 🐢

Issue #924, opened 2019-08-19
Combat rules doc 1 references old armor rules


Turtlelord26:
Getting shot section

PR #928, opened 2019-08-27
Delete Getting Shot section

Turtlelord26:
#924 
Used old rules. Also redundant with rules text in Shooting and Armor rules docs.


bleehu (approved):
the more I look at these rules, the more I think we need to rewrite the way we explain them.

Turtlelord26:
That sounds like something to talk about at a standup

Issue #923, opened 2019-08-14
add star strider effect to background blurb


bleehu:
the effect in the example is missing the effect from the star strider perk

PR #931, opened 2019-08-29
Star-Strider Patch

rosebrallier:
Re-worded perk mechanical description with input from @Turtlelord26 
Added perk description to background example re: Issue #923


bleehu (approved):
shipit!

PR #937, opened 2019-08-31
how did we not have lockpicks yet?

bleehu:
modified:   07_Items.txt


Turtlelord26 (approved):
Because no Rogue-equivalent class.

My only concern is that skeleton key means something specific for warded locks and is not related to a bump key. Maybe describe it as a cut-down key?

Also we need locks now.

bleehu:
Locks should be in GM guide, I think.

PR #940, opened 2019-09-05
Turtlelord26 species stat points patch

Turtlelord26:
Day Zero-ing on species, from informal gripes.

### DIFF
--- a/CharacterCreation/01_01_Species_Human.txt
+++ b/CharacterCreation/01_01_Species_Human.txt
@@ -77,7 +77,7 @@ upon.
 ===== Passives =====
 
 *  Gain 1 additional language and learn new languages at twice the normal rate.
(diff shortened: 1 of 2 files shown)


Reveraine (approved):
Yay patch! 
We may need to see how people feel about not having any difference in Stat point changes between species, but species need testing anyway, so that will come with that, I think.

bleehu (approved):
I think this hot-patch is necessary, but I'd like to move to re-instate having differing species vary by stat points. I think differing stats will make sense to players.

Turtlelord26:
Let's talk about that sometime. I haven't liked stat points from species because, functionally, they do very little because of how few rules we have for how you can allocate your starting stats (and going from 6 to 7 costs the same as 8 to 9).

Issue #741, opened 2018-12-24
Gripe - Backstab Feat


Reveraine:
Backstab Feat Level progression feels weak.

Turtlelord26:
Should I also mention that, despite being named back*stab*, it doesn't actually require the damage to be dealt with a melee weapon as currently worded? Going to add that as an ask to this issue.

PR #942, opened 2019-09-08
fixed backstab working for ranged attacks

bleehu:
modified:   03_Feats.txt

Should address issue #741


Reveraine (approved):
Very Nice! I like the simplification and the flavor text!

PR #943, opened 2019-09-08
added point-blank feat

bleehu:
modified:   03_Feats.txt

### DIFF
--- a/CharacterCreation/03_Feats.txt
+++ b/CharacterCreation/03_Feats.txt
@@ -351,6 +351,23 @@ buzzed or hammered, you deal extra damage equal to your strength modifier. These
 bonuses only apply to melee attacks. You also get a +10% on rolls to avoid 
 blacking out.  
 
+== Point Blank ==
+
+Prerequisites:
+
+*  Backstab feat
+*  Profeciency with Pistols or SMGs
+
+    It's hard to miss if you shoot from less than a foot away. Luckily, you're
+sneaky enough that you end up that close with some frequency. You tend to target
+the gaps in armor around the neck or eyes, sometimes adding a quick punch or 
+fast kick to help contort your target into a vulnerable stance. 
+
+    You can use backstab with suppressed CQB guns as long as you're in melee
+range. As long as the target has less than 80% armor coverage, your attack
+bypasses armor.
+
+
 == Point-man ==
 Prerequisites:
 
--- a/CharacterCreation/03_Feats.txt
+++ b/CharacterCreation/03_Feats.txt
@@ -51,10 +51,10 @@ Prerequisites:
 Prerequisites: 
 
 *  Level 3
-*  Assassin
+*  Assassin or Stealth skill 30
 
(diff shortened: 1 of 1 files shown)


Reveraine (commented):
I like this idea!
The wording is a bit weird, but I understand what it is saying.

Turtlelord26:
Gunslingers are sad they have to take backstab to get this

Also wildlovelies will break this.

There are suppressible cqb weapons that are not pistols or smgs.

Nonstatic armor flag. I think 80% is equivalent to 5 (doublecheck MH armor)

bleehu:
What is a suppressable CQB weapon that isn't a pistol or SMG?

Turtlelord26:
This guy
https://github.com/bleehu/Compound_X/blob/dev/Items/04_04_Carbines_AssaultRifles_%26_BattleRifles.txt#L82

bleehu:
Can I get a review though?

CharacterCreation/03_Feats.txt
```diff
@@ -351,6 +351,23 @@ buzzed or hammered, you deal extra damage equal to your strength modifier. These
 bonuses only apply to melee attacks. You also get a +10% on rolls to avoid 
 blacking out.  
 
+== Point Blank ==
+
+Prerequisites:
+
+*  Backstab feat
+*  Profeciency with Pistols or SMGs
```
Turtlelord26: sp: Proficiency

CharacterCreation/03_Feats.txt
```diff
@@ -351,6 +351,23 @@ buzzed or hammered, you deal extra damage equal to your strength modifier. These
...
+
+    It's hard to miss if you shoot from less than a foot away. Luckily, you're
+sneaky enough that you end up that close with some frequency. You tend to target
+the gaps in armor around the neck or eyes, sometimes adding a quick punch or 
+fast kick to help contort your target into a vulnerable stance. 
+
+    You can use backstab with suppressed CQB guns as long as you're in melee
```
Turtlelord26: I would suggest instead of 'suppressed CQB guns' `CQB guns with a suppressor attached`, to help out our hypothetical players who may be less familiar with firearms terminology.

CharacterCreation/03_Feats.txt
Turtlelord26: But also, since we require Proficiency to use a class of firearms at all, a proficiency prereq is a little redundant.

Turtlelord26 (changes requested):
This feat is by far most effective with SMGs, but has Backstab as a prerequisite. Since backstab requires the Assassin class, which can never get SMGs, this doesn't resolve. I'll call that a blocker.
Perhaps Snap Shot would be a better prerequisite for this feat? If so, I would spell out the bonus instead of saying 'use backstab'

Turtlelord26 (approved):
Still mildly wary of the CQB carbine but I'll wait for somebody to break it. Looks good everywhere else; particularly like the backstab pre-req change.

Issue #504, opened 2018-05-12
Feature Request: Laser gun attachments 


bleehu:
Especially barrel attachments; a lot of the existing attachments only make sense for ballistic weapons. There is also no way of silencing a laser weapon.

Turtlelord26:
And/or Flavor text on the ones we currently have to explain why they also work on plasma/laser

dfuchs2:
How much noise does a laser gun make? Is it a loud hum? A hissing noise as hot coolant gas is vented from the gun? 
I feel like addressing the noise created by the gun is the first step in determining a way to silence it.

Turtlelord26:
Such questions are contained in #273, which while greenlit has been sitting quietly in a state of neglect for months while we've been working on more fundamental systems.

bleehu:
A laser in real life that is powered by banks of capacitors makes a tell-tale electric "snap" noise. It seems fair to me to expect a similar thing to happen with normal laser weapons in CX. 

Assigning to Masterdiekillsu who it sounds like has some ideas.

PR #945, opened 2019-09-08
added laser dampening manifold attachment

bleehu:
modified:   04_00_WeaponAttachments.txt

should help address Issue #504

### DIFF
--- a/Items/04_00_WeaponAttachments.txt
+++ b/Items/04_00_WeaponAttachments.txt
@@ -291,6 +291,26 @@ coilguns, shotguns, or Machine Guns.
 Auto and Burst Miss Chance: -0.5 
 R3 Range: +8m
 
+== Capacitor Manifold $500 ==
+
+    This aluminum-composite microfiber shroud fits over the capacitor housing of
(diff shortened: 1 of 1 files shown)


Reveraine (commented):
I like this attachment!
I'm not sure the explanation for how lasers function is necessary here. It might be better suited to the laser weapons document, to keep the attachment document succinct. 

I'm also not sure that it makes sense to me. If the laser is not visible, why does the muzzle glow? I think it may ask more questions than it answers. (For a layman, specifically)

Turtlelord26:
Agree with above on prose. Mechanics seem good.

One note: shotgun fire should be spreading shot shotgun fire unless you wanted to include slugs.

bleehu:
lasers can't use slugs?

Turtlelord26:
Laser shotguns have spreading shot and single shot modes, like ballistic ones. Or at least they should.

bleehu:
I think in both spread and focus fire, a shotgun pushes the kind of quick burst of energy that is likely to cause a problem with an insulating manifold.

rosebrallier (approved):
Looks cool, and I like the writing. Have at it!

Issue #507, opened 2018-05-14
Gripe Proposal: Hip-fire feat


Turtlelord26:
Investigate the possibility for a feat that improves hip-firing accuracy at point-blank ranges. May not be available to all classes / with all weapons.

bleehu:
@Eryneril, want to take a crack at it?

PR #946, opened 2019-09-08
added below the belt feat

bleehu:
modified:   03_Feats.txt
	modified:   ../Combat Rules/2_Shooting.txt

A reduction from -3 to -2 might not sound like a big deal, but it can be combined with several attachments to get fairly scary hipfire synergies. The range limit on it I think balances that out well, and the CQB qualifier means you can't hip-fire well with an assault rifle, chain gun, or rocket launcher. Looking for sanity check and push back. Issue #507

### DIFF
--- a/CharacterCreation/03_Feats.txt
+++ b/CharacterCreation/03_Feats.txt
@@ -56,6 +56,22 @@ Prerequisites:
     Any damage dealt to an unaware opponent from behind deals an additional 
 10 + (Lvl * 2) Base damage.
 
+== Below the Belt ==
+
+Prerequisites:
+
+*  Luck 4
+*  Dex 6
+
+    You've got enough luck, courage, and precision to shoot in a general 
+direction without taking the time to aim properly. Your firearms teacher would
+be very disappointed at the lack of form, but if you don't get out of this 
+sticky situation then you won't get to hear from him anyway.
+
+    The penalty for hip firing with a CQB weapon is reduced by 1 in your R1 
+range bracket. For more information on hip firing, see the "Hip Firing" section
+of the Shooting rules in combat rules.
+
 == Bulltrue ==
 Prerequisites: 
 
--- a/CharacterCreation/03_Feats.txt
+++ b/CharacterCreation/03_Feats.txt
@@ -70,7 +70,7 @@ sticky situation then you won't get to hear from him anyway.
 
     The penalty for hip firing with a CQB weapon is reduced by 1 in your R1 
 range bracket. For more information on hip firing, see the "Hip Firing" section
-of the Shooting rules in combat rules.
+of the Shooting Rules document in the Combat Rules directory.
 
 == Bulltrue ==
 Prerequisites: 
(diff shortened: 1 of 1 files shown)


Reveraine (approved):
I like this feat.

Turtlelord26:
If Ralaak takes this feat, he has no penalty for hip firing half the time (in r1)
In depth review when I get home

bleehu:
No penalty for firing an SMG single-fire at point-blank range. I think that's balanced.

CharacterCreation/03_Feats.txt
```diff
@@ -56,6 +56,22 @@ Prerequisites:
...
+*  Dex 6
+
+    You've got enough luck, courage, and precision to shoot in a general 
+direction without taking the time to aim properly. Your firearms teacher would
+be very disappointed at the lack of form, but if you don't get out of this 
+sticky situation then you won't get to hear from him anyway.
+
+    The penalty for hip firing with a CQB weapon is reduced by 1 in your R1 
+range bracket. For more information on hip firing, see the "Hip Firing" section
+of the Shooting rules in combat rules.
```
rosebrallier: "Combat Rules" (capitalized since it's a document title)

rosebrallier (approved):
One grammar change. Otherwise I say shipit pending @Turtlelord26 approval

CharacterCreation/03_Feats.txt
```diff
@@ -70,7 +70,7 @@ sticky situation then you won't get to hear from him anyway.
 
     The penalty for hip firing with a CQB weapon is reduced by 1 in your R1 
 range bracket. For more information on hip firing, see the "Hip Firing" section
-of the Shooting rules in combat rules.
+of the Shooting Rules document in Combat Rules directory.
```
rosebrallier: "in the Combat"

Turtlelord26:
There are several SMGs with 4 (-1 per 5 DEX) auto accuracy, most of which all shoot 6 bullets per action. AS an example, the lv 10 laser smg: https://github.com/bleehu/Compound_X/blob/dev/Items/04_03_SMGs_%26_Shotguns.txt#L144
If I have 5 dex, it autofires with -3 accuracy.
Add a lv 10 attachment called Heavy Recoil Compensator and a Grip-pod and I'm down to minus 0 accuracy during autofire.
Or at level 1, an angled foregrip plus a heavy profile receiver will do the trick as well.


This may be more of a comment on the autofire penalty `+4 - 1 for every 5 Dex` or on how several different kinds of attachment can buff autofire accuracy.
I'm inclined to let the feat in for now and see if Ralaak breaks it in half.

However, we've still got our CQB carbine sneaking in.
https://github.com/bleehu/Compound_X/blob/dev/Items/04_04_Carbines_AssaultRifles_%26_BattleRifles.txt#L82

bleehu:
I want to capture some notes about various "hip fire builds" in design before I merge this. I think you've shown we've come right up to the edge of being Iver powered. It creates a pitfall for future design.

Issue #672, opened 2018-09-09
Gripe: Frag grenades do what?


bleehu:
The explination of grenades doing 3 or 2 wounds is unnecessarily complex. https://github.com/bleehu/Compound_X/blob/R_C/Items/04_01_Melee_%26_Thrown_Weapons.txt#L310. We should reword and/or rework this to make it easier to understand and quicker to play.

Turtlelord26:
Noting that this is a dupe of #493 part 5, but I'm going to close out the old issue and keep this around because it's much cleaner and everything else in 493 was handled.

PR #948, opened 2019-09-08
Mh/issue 672/frag grenades

bleehu:
Should address minor grammar and Issue #672

### DIFF
--- a/Items/06_Explosives.txt
+++ b/Items/06_Explosives.txt
@@ -79,12 +79,18 @@ Armor Piercing Level: 1
 Reflex Modifier: -10
 Reflex Save: 100
 
-    Each Target within the blast radius takes three shrapnel hits. 45 ballistic 
-damage per hit, automatically hits armor. Targets take full damage to the torso, 
-legs, and arms (if armor is pierced or not present): (3*45 damage)+2xBleed 
-within a 2.5m radius.Only legs and torso (2*45 damage)+1xBleed within a 5m 
-radius. At least one hit must damage health to cause Bleed. Recall that multiple 
-hits create separate wounds, which each must be healed separately.
++-------+------+----------------+----------------+
+| Range | Hits | Damage per Hit |  Bleed Tokens  |
++-------+------+----------------+----------------+
+| 2.5 m |   3  |   45 damage    | 2 bleed tokens | 
++-------+------+----------------+----------------+
+|  5 m  |   2  |   45 damage    | 1 bleed token  |
++-------+------+----------------+----------------+
+
+    Each Target within the blast radius takes hits accoding to the table above. 
+Grenade shrapnel automatically hits armor if it is present. At least one hit 
+must damage health to cause Bleed. Recall that multiple hits create separate 
+wounds, which each must be healed separately.
 
     Each character in the radius of a Frag Grenade's range when it goes off may
 make a Reflex Save of DC 100. If they succeed, they may dive prone up to their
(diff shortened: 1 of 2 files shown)


Items/06_Explosives.txt
```diff
@@ -70,16 +73,18 @@ investigation checks.
...
 Reflex Modifier: -10
 Reflex Save: 100
 
-    Each Target within the blast radius takes multiple hits. 45 ballistic 
+    Each Target within the blast radius takes three shrapnel hits. 45 ballistic 
 damage per hit, automatically hits armor. Targets take full damage to the torso, 
 legs, and arms (if armor is pierced or not present): (3*45 damage)+2xBleed 
 within a 2.5m radius.Only legs and torso (2*45 damage)+1xBleed within a 5m 
-radius. At least one hit must damage health to cause Bleed.
+radius. At least one hit must damage health to cause Bleed. Recall that multiple
```
Reveraine: I like the changes here that make the damage more obvious. I think that there is still some ambiguity in the 'at least one hit causes bleed', does the player choose which wounds have the bleed tokens attached? Are the 2 tokens attached to one wound or are they one token per wound, but only 2? It would make them a little more powerful, but it would make more sense to me to have a bleed token for each shrapnel wound, just for simplicity's sake.

Items/06_Explosives.txt
```diff
@@ -112,16 +118,18 @@ at the cost of one action on their next turn.
...
 Armor Piercing Level: 2
 Reflex Modifier: -10
 Reflex Save: 120
 
-    Each Target within the blast radius takes multiple hits. 50 ballistic 
+    Each Target within the blast radius takes three shrapnel hits. 50 ballistic 
 damage per hit, automatically hits armor. Targets take full damage to the torso, 
 legs, and arms (if armor is pierced or not present): (3*50 damage)+2xBleed 
 within a 3m radius. Only legs and torso (2*50 damage)+1xBleed within a 6m radius. 
-At least one hit must damage health to cause Bleed.
```
Reveraine, line 124: Same ambiguity as above.

bleehu:
I hadn't seen that ambiguity, and now that you've called it out, I'm not sure how it should work. If a rule has that level of confusion, it's unplayable and you should feel free to block the PR by requesting changes. :) 
Good catch! 
I'll think on it and figure out how it should work, then clarify it.

Reveraine:
I also noticed while sketching out item cards that the effect prose on the ECSC Smoke Grenade is a bit wordy, especially when it comes to whether or not someone can be seen well. It says something like everyone is in 80% stealth and needs to roll an 81% to hit. I feel like this can be condensed to "a DC of 80 is required to hit or see anyone in or through the cloud. 
As I type this, I also realize that 81 is no longer a valid roll to hit.

Items/06_Explosives.txt
Turtlelord26: I suggest phrasing like: the frag grenade applies 3 wounds, and if any damage health then the target gets 2 bleed tokens.

bleehu:
I've created a new issue for smoke grenades.

Items/06_Explosives.txt
```diff
@@ -70,16 +73,18 @@ investigation checks.
...
+Range: 10+[DEX or STR]*2 meters
 Armor Piercing Level: 1
 Reflex Modifier: -10
 Reflex Save: 100
 
-    Each Target within the blast radius takes multiple hits. 45 ballistic 
+    Each Target within the blast radius takes three shrapnel hits. 45 ballistic 
 damage per hit, automatically hits armor. Targets take full damage to the torso, 
 legs, and arms (if armor is pierced or not present): (3*45 damage)+2xBleed 
 within a 2.5m radius.Only legs and torso (2*45 damage)+1xBleed within a 5m
```
rosebrallier: Another thing for this frag: I'd lead with "If all three shrapnel hits pierce the target's armor and the target is within 2.5m of the explosion, the target takes wounds to the torso, arms, and legs for a total of:.... Within a 5m radius of the explosion, targets take two shrapnel hits. If both hits pierce the target;s armor, they take wounds to the legs and torso for a total of:.... If only one shrapnel hit pierces the target's armor, they take a single wound to their torso for a total of:.... Targets do not suffer any Bleed effect unless at least one hit deals damage to their health.

rosebrallier (commented):
We may want to streamline the "damage within such and such radius" description into a table. If we don't, I made a suggestion for the prose.

Turtlelord26:
Maybe it would be beneficial to codify that DoT tokens are separate from and not tied to Wounds in the combat rules - taking damage section.

bleehu:
We'll simplify the health system soon, so let's not work too hard on codifying new rules that are too complex already.

Issue #715, opened 2018-11-12
Gripe: Stat Modifier Table


Reveraine:
Create a Stat modifier table, to decrease math required to make a character.

PR #950, opened 2019-09-08
added stat modifier table

bleehu:
modified:   CharacterCreation/00_00_Creating_a_character.txt

Issue #715


bleehu:
Why do math? (whispers) "'cause it's fun?"

Turtlelord26:
Math Good!

To keep our rules dry, this should be here or in combat rules, but not both. I like here.

bleehu:
The 00_00_character_creation doc is explicitly an optional meta-doc that quick-references how to make a character. I think it's appropriate in this rare instance to not keep the rules perfectly dry because of that doc's meta trait. Open to further push-back. I'd like to keep our rules as dry as possible.

rosebrallier (approved):
I'm with @bleehu on this one; since we have a location for these on our new character sheet, it's best to keep them in the character creation quick reference.

Turtlelord26:
Is our character creation walkthrough document even something we want to classify as optional?  Our character-building process is not the simplest, perhaps we should consider it as equal to a rules document?

Posit the questions: is it reasonable for a new player to make a character without access to the character creation document or an official character sheet? Or even an experienced one?

Also open to push-back

bleehu:
I want to be able to make quick-reference documents that copy info from one part of the rules to another, so I don't want to set the precedent that we absolutely adhere to dry rules. For another example, I present the glossary.

Turtlelord26 (approved):

Issue #938, opened 2019-09-01
Gripe: Muscle Memory Wording


rosebrallier:
> Negate the pump of a shotgun, cocking of a revolver, or cycling of a bolt 
> when performing another action besides the aforementioned (besides firing).

What even is this
In [Muscle Memory](https://github.com/bleehu/Compound_X/blob/6117aea0b1452e5f6c587522910cd4ad2947718e/CharacterCreation/03_Feats.txt#L324) feat

Turtlelord26:
I think (?) it's trying to say something like you can pump a shotgun during a move action (or going prone, etc.)

bleehu:
Oh! That makes more sense! I wasn't playing this right!

PR #954, opened 2019-09-11
Reword Muscle Memory

Turtlelord26:
re #938

One minor functional change, the prerequisites are now And instead of Or. It made wording the feat easier, but I am open to push-back.


bleehu (approved):
Shipit!

Issue #959, opened 2019-09-21
Lvl 2-3 soldier build breaks static armor


Wildlovelies:
Lvl 2-3 depends on $$
Str 6
Per 8
R1:    R2:   R3:
15m.  60m.  91m
-3.      -3.     -1.    
Full auto 6 rounds per action standard ammo
75 dmg per shot, 350 damage per action.
Feats: snapshot, combat expertise guns, assault rifle(10)
Gun:ecsc mk15 (assault rifle) $1715
Attachments: 
heavy barrel $720
1.2 holo sight $800
Angled foregrip $600
Muzzle brake $500 
-1 str sling (forgot price)
Heavy receiver $1000

~$5300 plus armor.

Wildlovelies:
The heaviest armor only covers by 6, most cover 2-4  full auto and never worry about armor for the most part. So apl never matters as long as you are accurate enough

Wildlovelies:
With match grade ammo, and other sling and 8 str, you can subtract another -1 from all ranges. Making it -4 -4 -2

Wildlovelies:
To ensure perspective, this lvl 1 gun can kill lvl 20 enemies because apl doesnt matter if you are accurate enough...

Turtlelord26:
I'm only calculating a 0,0,2 miss chance spread on your creation (counting match-grade and accuracy sling). Possibly I'm missing something since I did it mostly in my head. Just to be sure though, did you account for not being able to reduce your autofire penalty below one?
In the same vein, overall miss chance before autofire penalty is still bounded from below at -2.
https://github.com/bleehu/Compound_X/blob/dev/Combat%20Rules/2_Shooting.txt#L20 
So you can never be autofiring with a miss chance below -1.

Leaving aside 100% coverage armor, which I recall we've considered unhealthy since 1source days, it has always been the case that a level 1 character can deal damage to an arbitrarily high-level enemy as long as they were 'accurate enough' (it just manifested as armor rolls instead of being baked into the attack roll).

A character with 0 Miss Chance in the relevant range bracket shooting a target in heavy armor would hit that armor on Acc 1-6 and bypass it on 7-10 assuming no cover bonuses applied. With light cover, 80% of those shots are being interdicted by armor. The other 20% of shots are expected to deal 90 damage. That's one action, so we're in agreement that this is concerning, because 2 actions for 180 damage through the heaviest armor in the game is lethal to all PC-sized health pools.
[...]

PR #962, opened 2019-09-25
removed $500 from soldier class

bleehu:
Should address some of the concerns from issue #959


Turtlelord26:
My immediate concern is that this is both not solving underlying issues of weapon-armor-attachments systems interactions and can't stop a soldier from putting this build together in glass cannon fashion by neglecting some armor pieces.

PR #970, opened 2020-03-25
added knife sheath

bleehu:
modified:   04_00_WeaponAttachments.txt

### DIFF
--- a/Items/04_00_WeaponAttachments.txt
+++ b/Items/04_00_WeaponAttachments.txt
@@ -522,7 +522,7 @@ conceal in a sleeve or boot. It holds up to three throwing knives in easy reach.
 
     Dexterity requirement: 7
     Knives thrown from the sheath may be drawn as a free action. Refilling the
-sheath takes twelve Actions. 
+sheath takes four Actions, or three if a slight of hand(Dex) check of DC 45. 
 
 
 ===== Level 5 Requirement =====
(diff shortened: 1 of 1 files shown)


bleehu:
At the request of Wildlovelies. The idea behind the long reload time is to avoid spamming knives.

Items/04_00_WeaponAttachments.txt
```diff
@@ -515,6 +515,16 @@ draw your primary as a free action. Holstering your weapon is a free action.
...
     Strength Requirement:  -1
 
+== Concealed Throwing Knife Sheath $340 ==
+
+    This simple polylon band fits nicely around a wrist or ankle. Its easy to 
+conceal in a sleeve or boot. It holds up to three throwing knives in easy reach.
+
+    Dexterity requirement: 7
+    Knives thrown from the sheath may be drawn as a free action. Refilling the
+sheath takes twelve Actions.
```
Turtlelord26: It still takes time to draw spare knives from wherever you have them stored. Three turns seems very excessive.
I would suggest targeting the refill time such that pulling knives from armor and putting them in (a pair of) sheaths takes a turn. Say 2 or 3 actions based on a reload DC.

If quick draw isn't rate limited for throwing weapons, that's a different problem with a different solution.

bleehu: My concern is that this could be used to bypass existing quickdraw limits.

bleehu:
@Turtlelord26 is that more reasonable?

Turtlelord26:
Yes, but I think I'm going to push for the reload check to help out weapon skills. Sleight of hand is already super useful. Call it a nonblocking request.

Turtlelord26 (approved):

PR #971, opened 2020-03-25
rate limited quickdraw

bleehu:
modified:   CharacterCreation/03_Feats.txt

### DIFF
--- a/CharacterCreation/03_Feats.txt
+++ b/CharacterCreation/03_Feats.txt
@@ -445,6 +445,7 @@ Prerequisites:
     As a free action, draw your chosen weapon(s) into firing position. If you 
 are holding something, you can drop it and still draw your weapon as a free 
 action. However, you might not want to drop anything glass. Or any explosives.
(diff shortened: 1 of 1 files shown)


Turtlelord26 (commented):
Honestly I would say 1 weapon per turn, and we can make an exception for specifically light throwing/melee weapons (maybe 3 for those?)

bleehu:
Two lets you draw with dual wield, and gets you close to three for light melee and throwing weapons, while maintaining simplicity.

Turtlelord26:
Ah, I'd forgotten about dual-wielding. I'm still leery of things like being able to quickdraw your LMG and then your Flamethrower on a strong back specialist. Is there a way to have the cake and eat it too?

Turtlelord26 (approved):
If you can't think of one, go ahead and put this in for now and we can revisit it later if it turns out to be a balance problem.

bleehu:
With 4 Actions a turn, shouldn't be a problem. That would be enough to aim each and fire only once.

Issue #967, opened 2020-03-07
thrown weapon accuracy mods too high


Wildlovelies:
== Gurber Throwing Knife $80 ==

Damage: 34 + 2*DEX mod Ballistic + 1xBleed
Size: Light
DEX Requirement: 5
Armor Piercing Level: 2

	Ranged Miss Chance

* 5 out to 7 meters
* 6 out to 15 meters

	May be used for melee strikes with a -1 Accuracy penalty. 
Can carry 2 per Secondary Ammo Slot.

having a base 5 miss without the ability to increase it besides perception (EG. no sights or attachments) add on armor and you base miss chance is about 7. even with 9 perception, you are still at a 6. if they are even in light cover that makes it and 8 or 9. makes it so that throwing knives can almost never hit or use its armor pierce ability like melee. I believe that we need to treat it more like a melee weapon and use weapon skill: thrown as part of the equation. because in real life that does make sense. a novice who has never thrown one before is not going to hit anything, someone who has been throwing all their life would be deadly. use similar aiming mechanics as melee, even if it is half of melee scaling, like thrown/20 round down

PR #972, opened 2020-03-25
radically reduced accuracies of thrown weapons

bleehu:

re Issue #967


Turtlelord26 (approved):
I hope we can find an equivalent to attachments for these, but this is excellent for now.

PR #973, opened 2020-03-28
	deleted:    CharacterCreation/09_Tactical Movement.txt

bleehu:
Is this even helping anyone? If not, lets trim the fat.


Turtlelord26 (commented):
Maybe not us, but it might be very useful to prospective players unfamiliar with squad-shooter mechanics and/or irl tactics, especially with a dash of our combat mechanical text added in. I think save it for the long haul.

bleehu:
You've convinced me. We'll keep it.

PR #976, opened 2020-03-28
added berserker class feat distictive style

bleehu:
modified:   CharacterCreation/02_Classes.txt


Turtlelord26 (approved):
I'm not sure if this survives a class feat restructure but it looks fun for now

Issue #932, opened 2019-08-29
Add tool sets to Profession perks


rosebrallier:
Tools of the trade are cool.

rosebrallier:
We've got clothing written in; still need tools. @bleehu can you coordinate Matt and Shrub to write them in?

PR #980, opened 2020-03-31
Tools from Backgrounds

rosebrallier:
Characters gain a tool set based on their Profession. Re Issue #932

### DIFF
--- a/CharacterCreation/04_Backgrounds.txt
+++ b/CharacterCreation/04_Backgrounds.txt
@@ -25,6 +25,7 @@ three categories. For example, if you chose "Trioch System", "Caveat Emptor" and
 Tunra language dialects and 6 skill in Insight at level 1 and a +10 bonus to skill 
 checks to move in microgragity.
 
+== Clothing ==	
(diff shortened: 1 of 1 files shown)


Turtlelord26 (approved):
Very hands-off approach. Totally on-board, but as a thought experiment, what would you think of putting a 'suggested tool set' on each professional background?

Slight concern that your section header Tools of the Trade shares its name with a professional background.

rosebrallier:
> Very hands-off approach. Totally on-board, but as a thought experiment, what would you think of putting a 'suggested tool set' on each professional background?
Have at it!
 
> Slight concern that your section header Tools of the Trade shares its name with a professional background.
Oh. Yep. I agree; no idea what else to title it with yet.

Turtlelord26:
> > Very hands-off approach. Totally on-board, but as a thought experiment, what would you think of putting a 'suggested tool set' on each professional background?
>
> Have at it!

On reflection, definitely a new issue to be considered in another PR.

Reveraine (approved):
What about leaving Tools of the Trade as-is here and changing the name of the professional experience?

PR #981, opened 2020-04-04
added Rhino feat

bleehu:
modified:   CharacterCreation/03_Feats.txt


Turtlelord26 (changes requested):
It should say that if an effect would reduce your speed below 2, it instead reduces it to 2. Prevents feel-bad from situations where an inflict 4 would be better than an inflict 5 but you only have 5.

Turtlelord26 (approved):

PR #1009, opened 2020-08-28
Jt backgrounds copy item tools assimilation

Turtlelord26:
Proposing assimilation of some tool-set-like things from the items document to the tool set document.


bleehu (approved):
Shipit!

